PR #30, opened 2014-11-04
Fixes #8266: Allow setting the number of Pulp workers to be used.

ehelms:
This also adds logic to set the Pulp workers based on processor count
but to cap the maximum at 8 unless overriden by a user.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -32,4 +32,28 @@ class pulp::params {
   $proxy_port     = undef
   $proxy_username = undef
   $proxy_password = undef
+
+  $num_workers = min($::processorcount, 8)
+
+  $osreleasemajor = regsubst($::operatingsystemrelease, '^(\d+)\..*$', '\1')
+
+  case $::operatingsystem {
+    'RedHat': {
+      case $osreleasemajor {
+        '6': {
+          $pulp_workers_template = 'upstart_pulp_workers'
+        }
+        default: {
+          $pulp_workers_template = 'systemd_pulp_workers'
+        }
+      }
+    }
+    'Fedora': {
+      $pulp_workers_template = 'systemd_pulp_workers'
+    }
+    default: {
+      fail("${::hostname}: This module does not support osfamily ${::operatingsystem}")
+    }
+  }
+
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,6 @@ rvm:
   - 2.1.0
 script: bundle exec rake test
 env:
-  - PUPPET_VERSION="~> 2.7.0"
   - PUPPET_VERSION="~> 3.2.0"
   - PUPPET_VERSION="~> 3.3.0"
   - PUPPET_VERSION="~> 3.4.0"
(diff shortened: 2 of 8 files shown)


templates/pulp_workers
```diff
@@ -0,0 +1,8 @@
+# Configuration file for Pulp's Celery workers
+
+# Define the number of worker nodes you wish to have here. This defaults to the number of processors
+# that are detected on the system if left commented here.
+PULP_CONCURRENCY=<%= @num_workers %>
+
+# Configure Python's encoding for writing all logs, stdout and stderr
+PYTHONIOENCODING="UTF-8"
```
jlsherrill: I believe the contents of this file depend on el6 vs el7:

https://github.com/pulp/pulp/blob/master/server/etc/default/systemd_pulp_workers
https://github.com/pulp/pulp/blob/master/server/etc/default/upstart_pulp_workers

On my el6 box, i have the other file.

ehelms:
@jlsherrill Updated to account for the different OS templates.

manifests/params.pp
```diff
@@ -31,4 +31,26 @@
   $proxy_port     = undef
   $proxy_username = undef
   $proxy_password = undef
+
+  $num_workers = min($::processorcount, 8)
+
+  case $::operatingsystem {
+    'RedHat': {
+      case $::operatingsystemmajrelease {
```
ekohl: Note this is a facter 1.7+ fact and thus unavailable on EPEL.

ehelms: Is there an equivalent?

ekohl: parsing operatingsystemrelease appears to be the most common method. We certainly use it all over the place in theforeman modules.

ehelms:
@ekohl @jlsherrill  Please review again.

manifests/params.pp
```diff
@@ -32,4 +32,28 @@
...
   $proxy_username = undef
   $proxy_password = undef
+
+  $num_workers = min($::processorcount, 8)
+
+  $osreleasemajor = regsubst($::operatingsystemrelease, '^(\d+)\..*$', '\1')
+
+  case $::operatingsystem {
+    'RedHat': {
+      case $::osreleasemajor {
```
ekohl: Since you're creating this var, it's not on the top scope. Should be ```$osreleasemajor```.

ehelms:
@ekohl updated

ekohl:
I'm surprised the ```@num_workers``` works in the templates, but if the tests pass :+1:

ehelms:
@ekohl I updated to use scope[] instead

ekohl:
@ehelms I don't think that's supported in older puppet versions and don't you need the absolute variable in newer versions?

ehelms:
I changed this to drop the 2.7 testing -- I can split that part out into another PR and go through a version rev cycle if people prefer. However, since we don't advertise or publish these officially I was less concerned. Further, I am not aware of any usage of this where 2.7 is supported (our installer requires 3+)

ekohl:
:+1: pending green tests

jlsherrill:
ACK from me

PR #32, opened 2015-01-07
change mongodb logpath

tmclaugh:
Change mongodb logpath from /var/lib/mongodb/mongodb.log to
/var/log/mongodb/mongodb.log.  Current results in a duplicate resource
error because the mongodb module attempts to create the logdir parent
which is the same as the dbpath.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -114,7 +114,7 @@ class pulp (
   }
   class { 'apache::mod::wsgi':} ~>
   class { 'mongodb':
-    logpath     => '/var/lib/mongodb/mongodb.log',
+    logpath     => '/var/log/mongodb/mongodb.log',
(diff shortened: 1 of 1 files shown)


tmclaugh:
Looks like the issue I ran into was due to a local change in puppetlabs-mongodb which I added to work around a bad package I was dealing with.  Even if not a bug, worth changing to logging to a standard place?

tmclaugh:
Local fix in my fork: https://github.com/hubspotdevops/puppetlabs-mongodb/commit/0cadd32df24ebe1c87a52abfd1c385a152298565

ehelms:
@tmclaugh This is no longer valid as we've changed the design around such that customization of Mongo should happen outside of the Pulp module itself.

PR #33, opened 2015-02-02
add initial admin support

cristi1979:
add initial admin support

### DIFF
--- a/manifests/admin.pp
+++ b/manifests/admin.pp
@@ -1,0 +1,99 @@
+#
+# == Class: pulp::admin
+#
+# Install and configure Pulp admin
+#
+# === Parameters:
+#
+# $version::                       pulp admin package version, it's passed to ensure parameter of package resource
+#                                  can be set to specific version number, 'latest', 'present' etc.
+#
+# $host::                          The pulp server hostname
+#
+# $port::                          The port providing the RESTful API
+#                                  type:integer
+#
+# $api_prefix::                    The REST API prefix.
+#
+# $verify_ssl::                    Set this to False to configure the client not to verify that the server's SSL cert is signed by
+#                                  a trusted authority
+#                                  type:boolean
+#
+# $ca_path::                       This is a path to a file of concatenated trusted CA certificates, or to a directory of trusted
+#                                  CA certificates (with openssl-style hashed symlinks, one certificate per file).
+#
+# $upload_chunk_size::             upload_chunk_size
+#                                  type:integer
+#
+# $role::                          The client role.
+#
+# $extensions_dir::                The location of admin client extensions.
+#
+# $id_cert_dir::                   The location of the directory where the Pulp user ID certificate is stored.
+#
+# $id_cert_filename::              The name of the file containing the PEM encoded client private key and X.509
+#                                  certificate. This file is downloaded and stored here during login.
+#
+# $upload_working_dir::            Directory where status files for in progress uploads will be stored
+#
+# $log_filename::                  The location of the admin client log file.
+#
+# $call_log_filename::             If present, the raw REST responses will be logged to the given file.
+#
+# $poll_frequency_in_seconds::     Number of seconds between requests for any operation that repeatedly polls
+#                                  the server for data.
+#                                  type:integer
+#
+# $enable_color::                  Set this to false to disable all color escape sequences
+#                                  type:boolean
+#
+# $wrap_to_terminal::              If wrap_to_terminal is true, any text wrapping will use the current width of
+#                                  the terminal. If false, the value in wrap_width is used.
+#                                  type:boolean
+#
+# $wrap_width::                    The number of characters written before wrapping to the next line.
+#                                  type:integer
+#
+# $puppet::                        Install puppet extension
+#                                  type:boolean
+#
+# $docker::                        Install puppet extension
+#                                  type:boolean
+#
+# $nodes::                         Install puppet extension
+#                                  type:boolean
+#
+# $python::                        Install puppet extension
+#                                  type:boolean
+#
+# $rpm::                           Install puppet extension
+#                                  type:boolean
+#
+class pulp::admin (
+  $version            = $pulp::admin::params::version,
+  $host               = $pulp::admin::params::host,
+  $port               = $pulp::admin::params::port,
+  $api_prefix         = $pulp::admin::params::api_prefix,
+  $verify_ssl         = $pulp::admin::params::verify_ssl,
+  $ca_path            = $pulp::admin::params::ca_path,
+  $upload_chunk_size  = $pulp::admin::params::upload_chunk_size,
+  $role               = $pulp::admin::params::role,
+  $extensions_dir     = $pulp::admin::params::extensions_dir,
+  $id_cert_dir        = $pulp::admin::params::id_cert_dir,
+  $id_cert_filename   = $pulp::admin::params::id_cert_filename,
+  $upload_working_dir = $pulp::admin::params::upload_working_dir,
+  $log_filename       = $pulp::admin::params::log_filename,
+  $call_log_filename  = $pulp::admin::params::call_log_filename,
+  $poll_frequency_in_seconds = $pulp::admin::params::poll_frequency_in_seconds,
+  $enable_color       = $pulp::admin::params::enable_color,
+  $wrap_to_terminal   = $pulp::admin::params::wrap_to_terminal,
+  $wrap_width         = $pulp::admin::params::wrap_width,
+  $puppet             = $pulp::admin::params::puppet,
+  $docker             = $pulp::admin::params::docker,
+  $nodes              = $pulp::admin::params::nodes,
+  $python             = $pulp::admin::params::python,
+  $rpm                = $pulp::admin::params::rpm,
+) inherits pulp::admin::params {
+  class { 'pulp::admin::install': } ~>
+  class { 'pulp::admin::config': }
+}
--- a/manifests/admin.pp
+++ b/manifests/admin.pp
@@ -69,6 +69,11 @@
 # $rpm::                           Install puppet extension
 #                                  type:boolean
 #
+# $puppet_upload_working_dir::     Directory where status files for in progress uploads will be stored
+#
+# $puppet_upload_chunk_size::      Maximum amount of data (in bytes) sent for an upload in a single request
+#                                  type:integer
+#
 class pulp::admin (
   $version            = $pulp::admin::params::version,
   $host               = $pulp::admin::params::host,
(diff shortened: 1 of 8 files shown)


metadata.json
```diff
@@ -9,26 +9,26 @@
...
-       "name": "theforeman-concat_native",
+       "name": "theforeman/concat_native",
        "version_requirement": ">= 1.3.0"
     },
     {
-       "name": "katello-qpid",
+       "name": "katello/qpid",
        "version_requirement": ">= 0.1.0"
     }
   ]
```
cristi1979: puppet still has this bug: https://tickets.puppetlabs.com/browse/FORGE-52

Meaning that "puppet module install" will fail to find any of the dependencies

manifests/admin.pp
```diff
@@ -0,0 +1,17 @@
...
+#
+class pulp::admin (
+  $version           = 'installed',
+  $host,
+  $port              = 443,
+  $api_prefix        = '/pulp/api',
+  $verify_ssl        = true,
+  $ca_path           = '/etc/pki/tls/certs/ca-bundle.crt',
+  $upload_chunk_size = '1048576',
+  $puppet            = false,) {
```
stbenjam: We try to keep class parameters in the params.pp style, these should move to `pulp::admin::params`

manifests/admin/install.pp
```diff
@@ -0,0 +1,8 @@
+# Pulp Admin Install Packages
+class pulp::admin::install {
+  package { ['pulp-admin-client',]: ensure => $pulp::admin::version }
+
+  if $pulp::admin::puppet {
+    package { ['pulp-puppet-admin-extensions',]: ensure => $pulp::admin::version }
+  }
+}
```
stbenjam: Nitpick, but my own preference, and the style in our modules tends to be without an array for single packages, and a new line after the resource title, e.g.

```
resource { 'title':
  param1 => value,
  param2 => value,
}
```

metadata.json
stbenjam: Thanks, didn't realize - we tend to not use puppet module install directly.  More info here, for some reason they seem unable to match the style here as everywhere else :-\

https://docs.puppetlabs.com/puppet/latest/reference/modules_publishing.html#dependencies-in-metadatajson

templates/etc/admin/admin.conf.erb
```diff
@@ -0,0 +1,87 @@
...
+#   CA certificates (with openssl-style hashed symlinks, one certificate per file).
+
+[server]
+host: <%= scope['pulp::admin::host'] %>
+port: <%= scope['pulp::admin::port'] %>
+api_prefix: <%= scope['pulp::admin::api_prefix'] %>
+verify_ssl: <%= scope['pulp::admin::verify_ssl'] %>
+ca_path: <%= scope['pulp::admin::ca_path'] %>
+upload_chunk_size: <%= scope['pulp::admin::upload_chunk_size'] %>
+
```
stbenjam: Should any of the other options in this file be configured by pulp admin?

cristi1979: I could add all options as parameters if you want.

stbenjam:
Thanks @cristi1979 :-)

 I'll let someone else who knows pulp than I do also review. 

Looks like tests are failing unrelated to your changes.  It's due to qpid module updates - in puppet-pulp here: https://github.com/Katello/puppet-pulp/blob/master/manifests/init.pp#L124, we are saying ssl true, but sending possibly no SSL certificates - puppet-qpid now validates that, I think we probably need to add some kind of $ssl parameter to the `pulp` class.

templates/etc/admin/admin.conf.erb
stbenjam: I think we don't need to expose everything, but if there's something generally useful... I don't know pulp well enough to say.

cristi1979:
I've add documentation and parametrize everything.

Regarding this:
# rsa_pub:
#   The pulp server public key used for authentication.

It doesn't seem to be used anywhere?

cristi1979:
I squashed all commits. Any other issues with the pull request?

templates/etc/admin/admin.conf.erb
```diff
@@ -0,0 +1,87 @@
+# The pulp server configuration
```
ehelms: I'd add an indicator that this file is managed by puppet at the top similar to -- https://github.com/theforeman/puppet-foreman/blob/master/templates/_header.erb

ehelms: I tend to prefer that templates just exist at the top level of the folder, i.e. `templates/admin.conf.erb`. I know we have some hold over from when this was refactored that still has some files under the "path" it would exist, but since this can change and there isn't an overload of templates, I'd prefer they sit under templates/

manifests/admin.pp
```diff
@@ -0,0 +1,99 @@
...
+  $poll_frequency_in_seconds = $pulp::admin::params::poll_frequency_in_seconds,
+  $enable_color       = $pulp::admin::params::enable_color,
+  $wrap_to_terminal   = $pulp::admin::params::wrap_to_terminal,
+  $wrap_width         = $pulp::admin::params::wrap_width,
+  $puppet             = $pulp::admin::params::puppet,
+  $docker             = $pulp::admin::params::docker,
+  $nodes              = $pulp::admin::params::nodes,
+  $python             = $pulp::admin::params::python,
+  $rpm                = $pulp::admin::params::rpm,
+) inherits pulp::admin::params {
```
ehelms: I am not seeing an admin params file in the PR?

ehelms:
It would be nice to see even a basic set of tests. We didn't do a good job of adding tests initially to our various puppet modules, but I want to see new stuff have at least a basic set of tests as we go forward.

templates/etc/admin/admin.conf.erb
stbenjam: It should be all or nothing though, if you want admin moved to the top level, everything else should be too

ehelms: I'm not disagreeing, sometimes I just stop trying to propagate it but I'll just do it all.

cristi1979:
tests fail in template admin.conf.erb for ruby 1.8.7 and PUPPET_VERSION="~> 2.7.0". Is this really important?

stbenjam:
@cristi1979 Considering 2.7 is EOL, I don't think so, I think we can use `scope[]`, @ehelms What do you think about dropping 2.7 support?

ekohl:
Considering 2.7 is still in EPEL5 and EPEL6, it's hard because I'm guessing a lot of sat6 customers are using it.

stbenjam:
No, we ship Puppet 3.6.2 with EL6 for Satellite.  For Katello, we have users use the Puppet labs repos, so they even get 3.7.x.

EL5 *clients* run 2.7, but they won't ever run Satellite.

ekohl:
That's good to know.

cristi1979:
I'm trying to create spec tests for the admin part, but it's not executed. I've just started looking over how to write those, so I'm not sure how to fix it. Currently it seems that it's only running this:

/home/travis/.rvm/rubies/ruby-1.8.7-p374/bin/ruby -S rspec spec/classes/coverage_spec.rb spec/classes/pulp_admin_spec.rb spec/classes/pulp_spec.rb

I would also like to run spec/acceptance/class_admin_spec.rb

Any help?

cristi1979:
I've added some spec tests now. for admin files and packages

spec/classes/pulp_admin_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+    context 'install class without parameters' do
+      it { should contain_class('pulp::admin::install') }
+      it { should contain_class('pulp::admin::config') }
+      it { should contain_class('pulp::admin::params') }
+
+      it { should contain_package('pulp-admin-client').with_ensure('installed') }
+
+      it 'should set admin.conf file' do
+        should contain_file('/etc/pulp/admin/admin.conf').
+        with_content(/^\[server\]$/).
```
ekohl: For readability I'd indent these with_ statements. For future reference you could also have a look at [verify_contents](https://github.com/puppetlabs/puppetlabs_spec_helper/blob/2c4550df345c40fb5b7f559b045bb03ee4e98611/lib/puppetlabs_spec_helper/module_spec_helper.rb#L9-L12):
```ruby
verify_contents(catalogue, '/etc/pulp/admin/admin.conf', ['line1', 'line2'])
```

Note the catalogue part is rspec-puppet 2.0, it was something else in rspec-puppet 1.x.

manifests/admin/config.pp
```diff
@@ -0,0 +1,14 @@
+# Pulp Admin Configuration
+class pulp::admin::config {
+  file { '/etc/pulp/admin/admin.conf':
+    ensure  => 'file',
+    content => template('pulp/admin.conf.erb'),
+  }
+
+  if $pulp::admin::puppet {
```
ekohl: If it's false, should it ensure that it's absent?

templates/admin_puppet.conf.erb
```diff
@@ -0,0 +1,8 @@
+### File managed with puppet ###
+[puppet]
+
+# Directory where status files for in progress uploads will be stored
+upload_working_dir = ~/.pulp/puppet-uploads
```
ekohl: Does it make sense to parameterize this value as well?

ekohl:
I think in general it looks good. Some smaller inline comments.

templates/admin_puppet.conf.erb
cristi1979: Let me know if you want this.

manifests/admin/config.pp
cristi1979: Should I check based on $version = absent/purged and remove the files in this case? From my point of view the file should remain on the system even if the package is uninstalled.

spec/classes/pulp_admin_spec.rb
cristi1979: I'm not sure what should be indent with "with_". Can you give me an example?

ekohl: I'd indent the with_ part so you can see it's part of the previous code block:
```ruby
should contain_file('/etc/pulp/admin/admin.conf').
    with_content(/^\[server\]$/).
```

manifests/admin/config.pp
ekohl: Then I'd leave it as it is now.

templates/admin_puppet.conf.erb
ekohl: I don't have a particular preference since I don't know pulp that well. It depends on how often the user would change this value.

cristi1979:
I parametrized admin_puppet.conf.erb also.

ekohl:
I'm not sure why the tests fail, but otherwise it looks good.

cristi1979:
fixed

ekohl:
:+1: as code review, didn't test it

stbenjam:
@cristi1979 Thanks for all the contributions, finally got around to testing and looks like it does what it says it does!

PR #34, opened 2015-02-03
fixes #9204 - fix failing tests due to qpid module's validations

stbenjam:
You can't by default send puppet-qpid `:ssl => true` and no certificates.  I would probably prefer to change ssl default to false, but that's an API change in the module.  Instead I set dummy paths which is probably the only option at the moment. 

When we deal with http://projects.theforeman.org/issues/9207 on this module, we can make the incompatible API change, bump the version, etc.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -50,6 +50,9 @@
 #
 # $reset_cache::                Boolean to flush the cache. Defaults to false
 #
+# $qpid_ssl::                   Enable SSL in qpid or not
+#                               type:boolean
+#
 # $qpid_ssl_cert_db             The location of the Qpid SSL cert database
 #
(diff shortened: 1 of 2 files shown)


ehelms:
ACK

ehelms:
We should honestly pull this out of the pulp module since qpid isn't required by pulp and we make use of qpid in other areas of the katello deployment.

stbenjam:
Hm, I thought pulp required it? Or at least the workers do, right?

qpid could go elsewhere, but also https://github.com/Katello/puppet-pulp/pull/35 adds more here.
The dispatcher is solely for pulp's purposes too, not sure where we should actually put it.  At the top level?

ehelms:
@stbenjam qpid is an option for Pulp, but not required. They can for example also used rabbitmq.

PR #35, opened 2015-02-04
refs #8175 - configure dispatch router on pulp and pulp nodes

stbenjam:
Requires https://github.com/Katello/puppet-capsule/pull/34


stbenjam:
Will only do this on the capsule side (https://github.com/Katello/puppet-capsule/pull/34), as this is a katello-specific thing.

PR #39, opened 2015-02-19
fixes #9479: fix usage with centos

dustints:


dustints:
deb users?

dustints:
same logic as before just inclusive of CentOS...
previously deb users were going to fall into the default/fail category.

ehelms:
Ack

ekohl:
:+1:

iNecas:
:+1: This also seems to fix the travis failings on all the puppet modules that include this one

Issue #36, opened 2015-02-16
Do not install qpid and mongodb from pulp


cristifalcas:
Using those classes inside pulp module stops us from using any custom parameters for them. Also they can't be used anywhere else in the catalog (for other modules).

If you agree, I could make a pull request to remove the declaration of qpid and mongodb from pulp.

Or, in order to keep the current functionality, a new class can be used (pulp::extra or something) that can be included in pulp::init based on a parameter that defaults to true.

Currently we can't install pulp using the upstream module because we need to move the mongodb dbpath to another location.

stbenjam:
Removing qpid makes sense, I think it should probably get moved over `puppet-capsule`.  Mongodb is a hard requirement of pulp, although I wonder if we shouldn't also have that in the capsule module, and find a stock mongodb module to set it up.

stbenjam:
On the other hand, I think the pulp module should be able to configure pulp in a usable state by itself, so maybe the mongodb section could be enabled/dislabed?

cristi1979:
Maybe just include mongodb/qpid and let is use the stock parameters?

PR #40, opened 2015-03-03
switch to include instead of declaring classes

cristi1979:
- also: align variables
- $ssl_ca_cert was defined 2 times
- add $qpid_user_groups as a variable

I don't know what to do about qpid declaration. Should I make it an include also, remove it?


cristi1979:
With the current definition of qpid, install fails:

    sudo -u apache pulp-manage-db
    Applying migration pulp.server.db.migrations.0009_qpid_queues failed.
    Halting migrations due to a migration failure.  See log for details.
    connection-forced: Connection must be encrypted.(320)

Error message from qpid:

    Mar 03 14:52:44 v-is-puppet-03.synygy.net qpidd[8484]: 2015-03-03 14:52:44 [Security] error Rejected un-encrypted connection.

So it looks like qpidd needs additional configuration (require-encryption=no).

I was thinking to make qpid an include also? It works with the default qpid configuration.

cristi1979:
OK, I've put include qpid also and removed the qpid variables. Also, in order to httpd to start you need to include ::apache::mod::ssl

manifests/init.pp
```diff
@@ -71,77 +62,43 @@
...
-  if (versioncmp($::mongodb_version, '2.6.5') >= 0) {
-    $mongodb_pidfilepath = '/var/run/mongodb/mongod.pid'
-  } else {
-    $mongodb_pidfilepath = '/var/run/mongodb/mongodb.pid'
-  }
+  include ::mongodb::globals
+  include ::mongodb::server
+  include ::mongodb::client
+
+  include qpid
```
ekohl: I'd use ```include ::qpid```

stbenjam:
This fixes https://github.com/Katello/puppet-pulp/issues/36, at least partially, right?

I'm OK with this change in general, but we need to make sure it doesn't break the katello-installer.

For qpid, I think it should be enableable with a boolean (qpid=true) and only include then -- because users may want to use rabbitmq, which pulp itself supports.

manifests/init.pp
```diff
@@ -71,77 +62,43 @@
...
   include ::apache
+  include ::apache::mod::wsgi
+  include ::apache::mod::ssl
 
-  if (versioncmp($::mongodb_version, '2.6.5') >= 0) {
-    $mongodb_pidfilepath = '/var/run/mongodb/mongod.pid'
-  } else {
-    $mongodb_pidfilepath = '/var/run/mongodb/mongodb.pid'
-  }
+  include ::mongodb::globals
```
ekohl: If you're not overriding any globals, there's no point in including it since ```mongodb::params``` inherits from it.

manifests/init.pp
```diff
@@ -71,77 +62,43 @@
...
+  include ::mongodb::client
+
+  include qpid
 
-  class { 'mongodb::globals':
-    version => $::mongodb_version, # taken from the custom facts
-  }
-  class { 'apache::mod::wsgi':} ~>
-  class { 'mongodb':
-    logpath     => '/var/lib/mongodb/mongodb.log',
```
ekohl: Why was this override here in the first place? Looks like a weird place for logs.

cristi1979:
- the template needed auth_enabled: true in order for consumers to work. I've updated the content of server.conf.erb with data from the file installed by the rpm

ehelms:
@cristi1979 With this change, users are required to declare and configure qpid outside of the pulp module?

cristi1979:
qpid is included by default, so there is no need to add it externally.

This is how I test it on 2.5.3 and update it to 2.6 beta:

      class { 'pulp':
        default_password => 'admin',
        version          => 'latest',
      }

      class { 'pulp::admin':
        verify_ssl => false,
        version    => 'latest',
      }

      class { 'pulp::consumer':
        verify_ssl => false,
        version    => 'latest',
      }

Tests performed:

    pulp-admin login -u admin -p admin
    pulp-admin rpm repo create --repo-id=testes --relative-url=testes --feed=https://hostname/nada
    pulp-admin rpm repo delete --repo-id=testes
    pulp-admin rpm repo create --repo-id=testes --relative-url=testes --feed=https://hostname/nada
    pulp-admin rpm repo create --repo-id=pulp_beta --relative-url=pulp_beta --feed=https://repos.fedorapeople.org/repos/pulp/pulp/v2/stable/7Server/x86_64/
    pulp-admin rpm repo sync run --repo-id=pulp_beta
    pulp-admin rpm repo update --repo-id pulp_beta --serve-http=true
    pulp-admin rpm repo sync status --repo-id pulp_beta

    pulp-consumer -u admin -p admin register --consumer-id $(hostname)
    pulp-consumer rpm bind --repo-id=pulp_beta
    pulp-consumer rpm bind --repo-id=testes
    pulp-consumer rpm unbind --repo-id=testes
    yum remove atop -y
    pulp-admin rpm consumer package install run --consumer-id $(hostname) --name atop
    pulp-admin tasks list | grep "State:"
    pulp-consumer status
    pulp-admin consumer list

    pulp-consumer rpm unbind --repo-id=pulp_beta
    pulp-consumer unregister
    pulp-admin rpm repo delete --repo-id=testes
    pulp-admin rpm repo delete --repo-id=pulp_beta

ehelms:
What I mean is the lines - https://github.com/Katello/puppet-pulp/pull/40/files#diff-60ae41fd0a31977447947f59940ee9a4L130 appear to be removed, which removes the configuration parameters we previously passed for qpid.

cristi1979:
Yes, qpid will be installed and used by default whitout ssl. In order to use ssl, qpid module will have to be declared separately. This is also the case for pulp, that will require extra parameters. 

I didn't test this yet in order to see if it will work. I have to research this and how qpid with ssl works, so it may take some extra days.

I will do so in the next days, when my time will permit.

cristi1979:
qpid + pulp (on same host) work if I use the following definitions. Certs where created using: /bin/echo | /bin/pulp-qpid-ssl-cfg.

      class { 'qpid':
        ssl                    => true,
        ssl_cert_db            => '/etc/pki/pulp/qpid/nss',
        ssl_cert_password_file => '/etc/pki/pulp/qpid/nss/password',
        ssl_cert_name          => 'broker',
        #    ssl_require_client_auth => false,
        user_groups            => [],
      # auth => true,
      }
    
      class { 'pulp':
        default_password      => 'admin',
        version               => 'latest',
        ssl_enforce_verify    => false,
        enable_puppet         => true,
        messaging_url         => "ssl://${::fqdn}:5671",
        messaging_ca_cert     => '/etc/pki/pulp/qpid/ca.crt',
        messaging_client_cert => '/etc/pki/pulp/qpid/client.crt',
        broker_url            => "qpid://${::fqdn}:5671",
        broker_use_ssl        => true,
      }
    
      class { 'pulp::admin':
        verify_ssl    => false,
        version       => 'latest',
        enable_puppet => true,
      }
    
      class { 'pulp::consumer':
        verify_ssl           => false,
        version              => 'latest',
        enable_puppet        => true,
        messaging_scheme     => 'ssl',
        messaging_host       => $::fqdn,
        messaging_port       => 5671,
        messaging_cacert     => '/etc/pki/pulp/qpid/ca.crt',
        messaging_clientcert => '/etc/pki/pulp/qpid/client.crt',
      }

cristi1979:
About the above: it works, but I have to find a simple solution refresh Exec['migrate_pulp_db'] after config changes and after packages installation/updates

cristi1979:
I would like to make a new PR with 2 new parameters: $db_manage and $broker_manage. But it includes the changes from this PR. Should I wait for this one to be accepted or make the new PR anyway?

cristi1979:
- move to includes
- added parameters and documentation for server.conf
- added $pulp::db_manage and $broker_manage in order to decide if the broker and db are local or remote
- moved migrate_pulp_db exec to a different class, because it needs to be run after every config change (qpid+ssl for example)
- parameterize the server.conf file

cristi1979:
Should I close this PR and open smaller ones? Based on the current master?

cristi1979:
closing in order to make smaller PRs

PR #41, opened 2015-03-03
add consumer

cristi1979:
Add consumer configuration.

Should I add the agent installation/configuration here also? Or in a separate class "pulp:agent"?

### DIFF
--- a/manifests/consumer/config.pp
+++ b/manifests/consumer/config.pp
@@ -1,0 +1,54 @@
+# Pulp Consumer Configuration
+class pulp::consumer::config {
+  file { '/etc/pulp/consumer/consumer.conf':
+    ensure  => 'file',
+    content => template('pulp/consumer.conf.erb'),
+    owner   => 'root',
+    group   => 'root',
+    mode    => '0644',
+  }
+
+  if $pulp::consumer::enable_rpm {
+    file { '/etc/pulp/agent/conf.d/bind.conf':
+      ensure  => 'file',
+      content => template('pulp/agent_bind.conf.erb'),
+      owner   => 'root',
+      group   => 'root',
+      mode    => '0644',
+    }
+
+    file { '/etc/pulp/agent/conf.d/linux.conf':
+      ensure  => 'file',
+      content => template('pulp/agent_linux.conf.erb'),
+      owner   => 'root',
+      group   => 'root',
+      mode    => '0644',
+    }
+
+    file { '/etc/pulp/agent/conf.d/rpm.conf':
+      ensure  => 'file',
+      content => template('pulp/agent_rpm.conf.erb'),
+      owner   => 'root',
+      group   => 'root',
+      mode    => '0644',
+    }
+  }
+
+  if $pulp::consumer::enable_puppet {
+    file { '/etc/pulp/agent/conf.d/puppet_bind.conf':
+      ensure  => 'file',
+      content => template('pulp/agent_puppet_bind.conf.erb'),
+      owner   => 'root',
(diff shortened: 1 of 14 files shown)


cristi1979:
- added a provider to register hosts
- also added agent configuration

templates/consumer.conf.erb
```diff
@@ -0,0 +1,172 @@
...
+#   WARNING: Setting this to False is not safe if you wish to use Pulp with real passwords or
+#            any other private information. It is strongly recommended to keep this set to True, and
+#            to use certificates that are signed by a trusted authority on the web server that
+#            serves Pulp.
+# ca_path:
+#   This is a path to a file of concatenated trusted CA certificates, or to a directory of trusted
+#   CA certificates (with openssl-style hashed symlinks, one certificate per file).
+
+[server]
+host: <%= scope.lookupvar('pulp::consumer::host') %>
```
ehelms: Prefer `@` syntax or `scope[]`

ehelms:
Some tests and example usage would be a nice addition.

templates/consumer.conf.erb
cristi1979: puppet 2.7.0 doesn't support scope[] from what I know, so the tests will fail until that version is removed. Is this ok with you?

ehelms: @cristi1979 do you require 2.7 (since you are a prime user of this stand alone) ? We (the Katello team) have been removing 2.7 support from our other modules due to this (the scope[] support) and because our katello-installer doesn't support 2.7 anyway.

cristi1979: No, we're using 3.x. Just that the spec tests include tests for puppet 2.7, so travis will fail until they are removed.

cristi1979:
tests added

cristi1979:
moved from scope.lookupvar to scope

ehelms:
@cristi1979 do you have a preference for the order of these PRs? I have tagged and branched 0.1.0 with the intent that we will bring in your changes and consider them a 1.0.0 release due to the scale and breaking nature of the changes.

cristi1979:
I really don't care in what order the PRs are accepted. I'm trying to make my changes in different PRs, but I may have some changes on my branch that are not pushed yet, so I'm waiting for the bulk of them to get in and make new PRs after that.

Also my tests are not comprehensive, so I wouldn't desire a 1.0 version until some other people test their workflow.

cristi1979:
Also, I tried to make my changes independent of the current status, but that wasn't the case unfortunately. In order to implement some of them (the switch to include classes especially) I had to push changes in the default server template and in the config class. If those PRs are accepted, I would like to make the server template configurable also.

manifests/consumer/config.pp
```diff
@@ -0,0 +1,36 @@
+# Pulp Consumer Configuration
+class pulp::consumer::config {
+  file { '/etc/pulp/consumer/consumer.conf':
```
ehelms: What about setting owner/group/mode on these files? We have tended toward that as a best practice.

ehelms:
Small comment, and if you will squash to a single commit I'll merge this today.

manifests/consumer/config.pp
cristi1979: Who should own the files? root?

ehelms: Yea, 'root'

cristi1979: already done

ehelms:
ACK - thanks @cristi1979 !

PR #42, opened 2015-03-04
fix services on el7

cristi1979:
This is probably a bug in the rpms, but on el7 the services don't start until I issue "systemctl daemon-reload"


ehelms:
Would you mind also filling a bug with Pulp (https://pulp.plan.io/) so that we could remove this at a later date?

cristi1979:
https://pulp.plan.io/issues/742

cristi1979:
Any issues with this?

ehelms:
ACK - thanks @cristi1979

PR #44, opened 2015-03-04
add ssl config

cristi1979:
- the rest of the changes are from Geppetto autoformat


manifests/config.pp
```diff
@@ -109,32 +124,30 @@
 
   file { '/etc/default/pulp_workers':
     ensure  => file,
-    content => template("pulp/${pulp::pulp_workers_template}"),
+    content => template("pulp/${pulp::params::pulp_workers_template}"),
```
ekohl: This doesn't seem right.

manifests/config.pp
```diff
@@ -109,32 +124,30 @@
...
   if $pulp::reset_cache {
-    exec {'reset_pulp_cache':
+    exec { 'reset_pulp_cache':
       command => 'rm -rf /var/lib/pulp/packages/*',
       path    => '/sbin:/bin:/usr/bin',
       before  => Exec['migrate_pulp_db'],
-      require => [
-        File['/var/lib/pulp/packages'],
-        ],
+      require => [File['/var/lib/pulp/packages'],],
```
ekohl: Might as well change it to ```require => File['/var/lib/pulp/packages'],``` then

cristi1979:
- this configuration replaces pulp.conf.erb
- also adds firewall rules

ekohl:
In general it's uncommon to set firewall rules in a module. It's usually done in profiles and I'd certainly think it needs to be in a separate PR.

metadata.json
```diff
@@ -16,7 +16,6 @@
        "name": "puppetlabs/apache",
        "version_requirement": ">= 1.0.0 < 2.0.0"
     },
-    {
```
ekohl: Doesn't this break the json file?

manifests/config.pp
```diff
@@ -22,15 +21,72 @@
     mode    => '0600',
   }
 
-  file {'/etc/httpd/conf.d/pulp.conf':
```
ekohl: If you're moving to using ```apache::vhost``` (which I think is a good move), then you should remove the template (```templates/etc/httpd/conf.d/pulp.conf.erb```) as well.

manifests/config.pp
```diff
@@ -22,15 +21,72 @@
...
+    priority            => '05',
+    docroot             => '/srv/pulp',
+    port                => 80,
+    servername          => [$::fqdn],
+    serveraliases       => [$::hostname],
+    additional_includes => '/etc/pulp/vhosts80/*.conf',
+    add_default_charset => 'UTF-8',
+  }
+
+  apache::vhost {'pulp-https':
```
ekohl: I can't see you setting the ```WSGISocketPrefix```. Maybe this is done while configuring the module, but it's important to note.

manifests/config.pp
```diff
@@ -22,15 +21,72 @@
...
   }
 
-  file {'/etc/httpd/conf.d/pulp.conf':
+  apache::vhost { 'pulp-http':
+    priority            => '05',
+    docroot             => '/srv/pulp',
+    port                => 80,
+    servername          => [$::fqdn],
+    serveraliases       => [$::hostname],
+    additional_includes => '/etc/pulp/vhosts80/*.conf',
```
ekohl: This is only done for Fedora 18, which maybe should be dropped. @ehelms?

manifests/config.pp
```diff
@@ -22,15 +21,72 @@
...
+        'path'     => 'webservices.wsgi',
+        'provider' => 'files',
+      },
+      {
+        'path'     => '/srv/pulp',
+        'provider' => 'directory',
+      }
+      ],
+    options                    => ['SymLinksIfOwnerMatch'],
+    custom_fragment            => '  SSLInsecureRenegotiation on'
```
ekohl: The comment (allow older yum clients to connect, see bz 647828) is helpful to understand why it's needed and I'd at least leave it here.

manifests/config.pp
cristi1979: WSGISocketPrefix is set from apache::mod::wsgi class, so I didn't thought it's necessary to set it explicitly.

cristi1979:
- removed firewall rules
- removed old fact (mongodb_version)
- fix apache definition

.fixtures.yml
```diff
@@ -2,6 +2,7 @@ fixtures:
   repositories:
     stdlib:        "git://github.com/puppetlabs/puppetlabs-stdlib.git"
     foreman:       "git://github.com/theforeman/puppet-foreman.git"
+    firewall:      "git://github.com/puppetlabs/puppetlabs-firewall"
```
ekohl: This is no longer needed.

lib/facter/facts.rb
```diff
@@ -1,16 +0,0 @@
-Facter.add(:mongodb_version) do
```
ehelms: Is this removal directly related to this change?

manifests/config.pp
```diff
@@ -22,15 +21,72 @@
...
+    aliases                    => [{
+        alias           => '/pulp/static',
+        path            => '/var/lib/pulp/static',
+        options         => ['Indexes'],
+        custom_fragment => 'SSLRequireSSL'
+      }
+      ],
+    options                    => ['SymLinksIfOwnerMatch'],
+    custom_fragment            => '# allow older yum clients to connect, see bz 647828
+  SSLInsecureRenegotiation on'
```
ehelms: LInes 69-78 seem off indentation wise.

cristi1979:
I'm not sure how to push some other changes, because the files are already modified by other PRs.

I would like to add 2 new parameters:

     $ssl_enforce_verify = true
     $enable_http = false

$ssl_enforce_verify should allow https access to the repos without requiring authenticating with the certificates (browsers and yum without ssl certificates).

cristi1979:
http access is also required for puppet repos, at least.

manifests/config.pp
ehelms: Agreed, we should avoid the use of params

manifests/config.pp
```diff
@@ -22,15 +21,72 @@
...
   }
 
-  file {'/etc/httpd/conf.d/pulp.conf':
+  apache::vhost { 'pulp-https':
+    priority                   => '05',
+    docroot                    => '/srv/pulp',
+    port                       => 443,
+    servername                 => [$::fqdn],
+    serveraliases              => [$::hostname],
+    add_default_charset        => 'UTF-8',
```
ehelms: I am getting a `Invalid parameter add_default_charset` when testing this.

cristi1979: I see it was added in apache 1.2.0. Should I remove it?

manifests/config.pp
cristi1979: Sorry about that, I will remove params.

manifests/config.pp
ehelms: If we keep it, we'd need to bump the metadata requirement, however, due to how we use this module currently I'd prefer to stay 1.1.0 compatible.

cristi1979:
rebased on the original config.pp

manifests/config.pp
```diff
@@ -14,12 +14,58 @@
...
+        custom_fragment => 'SSLRequireSSL'
+      }
+    ],
+    options             => ['SymLinksIfOwnerMatch'],
+    custom_fragment     => '# allow older yum clients to connect, see bz 647828
+  SSLInsecureRenegotiation on
+  AddDefaultCharset UTF-8'
   }
 
   file {'/etc/httpd/conf.d/pulp.conf':
```
ehelms: Looks like this needs to be removed.

manifests/config.pp
```diff
@@ -14,12 +14,58 @@
     mode   => '0755';
   }
 
-  file {'/etc/pulp/server.conf':
```
ehelms: DId you mean to remove this one or the one I pointed out below?

manifests/config.pp
```diff
@@ -14,12 +14,58 @@
...
-    ensure  => file,
-    content => template('pulp/etc/pulp/server.conf.erb'),
-    owner   => 'apache',
-    group   => 'apache',
-    mode    => '0600',
+  apache::vhost { 'pulp-https':
```
ehelms: What do you think about putting this in it's own class? `pulp::apache`

ehelms:
@cristi1979 Apologies for the long delay in getting back to these, releases and other things got me backed up. Hopefully you haven't given up on these; however, if you need someone to take them over for you please let us know.

cristi1979:
closing. I will do a new PR

PR #45, opened 2015-03-07
use preffix enable_ for extra packages

cristi1979:
- also $host = 'localhost' is always wrong, so use $::fqdn instead
- enable default rpm plugin

### DIFF
--- a/manifests/admin.pp
+++ b/manifests/admin.pp
@@ -54,19 +54,19 @@
 # $wrap_width::                    The number of characters written before wrapping to the next line.
 #                                  type:integer
 #
-# $puppet::                        Install puppet extension
+# $enable_puppet::                 Install puppet extension
 #                                  type:boolean
 #
-# $docker::                        Install puppet extension
+# $enable_docker::                 Install puppet extension
 #                                  type:boolean
 #
-# $nodes::                         Install puppet extension
+# $enable_nodes::                  Install puppet extension
 #                                  type:boolean
 #
-# $python::                        Install puppet extension
+# $enable_python::                 Install puppet extension
 #                                  type:boolean
 #
-# $rpm::                           Install puppet extension
+# $enable_rpm::                    Install puppet extension
 #                                  type:boolean
 #
 # $puppet_upload_working_dir::     Directory where status files for in progress uploads will be stored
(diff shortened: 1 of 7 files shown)


manifests/admin.pp
```diff
@@ -93,14 +93,13 @@
...
-  $rpm                = $pulp::admin::params::rpm,
+  $enable_puppet      = $pulp::admin::params::puppet,
+  $enable_docker      = $pulp::admin::params::docker,
+  $enable_nodes       = $pulp::admin::params::nodes,
+  $enable_python      = $pulp::admin::params::python,
+  $enable_rpm         = $pulp::admin::params::rpm,
   $puppet_upload_working_dir = $pulp::admin::params::puppet_upload_working_dir,
-  $puppet_upload_chunk_size  = $pulp::admin::params::puppet_upload_chunk_size,
-) inherits pulp::admin::params {
+  $puppet_upload_chunk_size  = $pulp::admin::params::puppet_upload_chunk_size,) inherits pulp::admin::params {
```
ehelms: Why move this closing paren and inherits to the same line as the last param?

cristi1979:
fixed geppetto autoformat

ehelms:
This is looking good - mind squashing commits?

cristi1979:
- squashed
- I also made enable_rpm = true by default

manifests/admin.pp
```diff
@@ -93,14 +93,14 @@
...
-  $puppet             = $pulp::admin::params::puppet,
-  $docker             = $pulp::admin::params::docker,
-  $nodes              = $pulp::admin::params::nodes,
-  $python             = $pulp::admin::params::python,
-  $rpm                = $pulp::admin::params::rpm,
+  $enable_puppet      = $pulp::admin::params::puppet,
+  $enable_docker      = $pulp::admin::params::docker,
+  $enable_nodes       = $pulp::admin::params::nodes,
+  $enable_python      = $pulp::admin::params::python,
+  $enable_rpm         = $pulp::admin::params::rpm,
```
ehelms: Any reason for the parameters and default params to mimic each other in name?

manifests/admin.pp
```diff
@@ -93,14 +93,14 @@
...
-  $puppet             = $pulp::admin::params::puppet,
-  $docker             = $pulp::admin::params::docker,
-  $nodes              = $pulp::admin::params::nodes,
-  $python             = $pulp::admin::params::python,
-  $rpm                = $pulp::admin::params::rpm,
+  $enable_puppet      = $pulp::admin::params::enable_puppet,
+  $enable_docker      = $pulp::admin::params::enable_docker,
+  $enable_nodes       = $pulp::admin::params::enable_nodes,
+  $enable_python      = $pulp::admin::params::enable_python,
+  $enable_rpm         = $pulp::admin::params::enable_rpm,
```
ehelms: Not a requirement for the PR, but you cold consider adding `validate_bool` for each to enforce the type

cristi1979:
I've also added 2 new packages to the puppet installation: 'pulp-puppet-tools', 'python-pulp-puppet-common'. The second one is required because of bug 1105673

cristi1979:
all tests for ruby 1.8.7 failed with an exception. Since I couldn't find how to restart the travis jobs, I will close/reopen this PR.

ekohl:
Test failures are due to https://github.com/rspec/rspec-core/issues/1864

stbenjam:
https://github.com/Katello/puppet-pulp/pull/47 fixes that

ekohl:
Nitpick: it's a prefix, not a suffix.

ehelms:
This looks good to me. @ekohl what say ye

manifests/admin/params.pp
```diff
@@ -1,7 +1,7 @@
 # Pulp Admin Params
 class pulp::admin::params {
   $version            = 'installed'
-  $host               = 'localhost'
+  $host               = $::fqdn
```
ekohl: This change appears to be unrelated and I don't know pulp well enough to know if it makes a difference.

ekohl:
I think it looks good. This does mean a major version bump of the module, but that can be ok. I don't know if a version shipped in the installer with these parameters, but that may also be worth considering and testing how it'll work for upgrading users.

cristi1979:
Template was changed to use scope[] instead of scope.lookupvar

ehelms:
@ekohl We cut a stable branch of 0.1 before these changes so that we could bring them all in and eventually release a 1.0.0. Does that sounds reasonable? Anything else you'd suggest?

ekohl:
@ehelms given you're in the 0.x series, I think that's fine. People should know a 0.x version can break the API with every release.

ehelms:
ACK - thanks @cristi1979

PR #48, opened 2015-03-14
Qpid+ssl with puppet certificates

cristi1979:
Example of using puppet certificates for qpid+ssl connectivity


cristi1979:
I was thinking to put this instead in an examples dir? On how to use pulp with ssl connectivity to broker and mongodb?

PR #50, opened 2015-03-14
Migrate db

cristi1979:
erun migrate db on package install/upgrade and config changes (qpid ssl needs it)

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -134,15 +134,6 @@ class pulp::config {
     }
   }
 
-  exec {'migrate_pulp_db':
-    command   => 'pulp-manage-db && touch /var/lib/pulp/init.flag',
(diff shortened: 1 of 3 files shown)


ehelms:
This comment plays into the switch to includes you have done elsewhere, but since it is a design related comment I'll post here. What we have done in other modules, and I think would be nice ot have here is to allow the module to declare and setup a mongodb. Since Pulp requires one to function, users should be able to pass a `manage_db` flag that will then have the module setup a database for you. If the flag is false, then it is expected that the user is providing the mongo database with connection credentials passed into the module. This would be done, including the migrate, in a `database.pp` class. An example of what I mean -- https://github.com/theforeman/puppet-foreman/blob/master/manifests/database.pp#L3

cristi1979:
closing in order to implement the new database class

PR #46, opened 2015-03-11
parameteryze installed packages

cristi1979:
Unfortunately changes from other PRs are included.

Can I wait for the other PRs to be accepted (rejected) and do a rebase based on that, or should I do the changes on the master branch?

PR #51, opened 2015-03-15
New server

cristi1979:
This is a rewrite of the server part.

Since pulp-agent is used by both consumers and child nodes, I moved it to a separate class which is included in case a child node is installed.

Template files from "etc/pulp" have been moved to root of template dir.

Child class has been removed.

Also, qpid and mongodb are installed only if the corresponding parameters are set to true: db_manage, broker_manage.

It replaces #44, #46, #40.


ehelms:
This is a lot to try to take in at once and glancing over it I'd think it might be easier to review and check if broken down into smaller changes. I could see the db_manage and broker_manage being their own separate changes, which with the addition of tests, would give me a better feeling of reliability on the changes. The moving of templates to top level could be an easy stand-alone change to review and then  the change to the child handling.

PR #52, opened 2015-03-15
move pulp agent to external class

cristi1979:


manifests/agent.pp
```diff
@@ -0,0 +1,18 @@
+#
+# == Class: pulp::agent
+#
+# Install and configure Pulp agent
+#
+class pulp::agent ($version) {
```
ehelms: Is this supposed to be able to support a stand-alone agent on a consumer? Looking below at the checks on `pulp::config` and `server.conf` it would seem this wouldn't work on a pure consumer.

cristi1979:
switched to if !defined(Class['pulp::agent']) in order to install the agent. This was done because the agent can be defined by the consumer or by a child node.

PR #53, opened 2015-05-05
Fixes #10385: Omit errors from yum info when acquiring version.

ehelms:
Newer mongodb module from Puppetlabs correctly handles setting the
pidfile and log paths.

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -3,7 +3,9 @@ fixtures:
     stdlib:        "git://github.com/puppetlabs/puppetlabs-stdlib.git"
     foreman:       "git://github.com/theforeman/puppet-foreman.git"
     apache:        "git://github.com/puppetlabs/puppetlabs-apache.git"
-    concat:        "git://github.com/puppetlabs/puppetlabs-concat"
+    concat:
+      repo:   "git://github.com/puppetlabs/puppetlabs-concat"
+      branch: "1.2.x"
     concat_native: "git://github.com/theforeman/puppet-concat"
     mongodb:       "git://github.com/puppetlabs/puppetlabs-mongodb.git"
     qpid:          "git://github.com/katello/puppet-qpid.git"
--- a/lib/facter/facts.rb
+++ b/lib/facter/facts.rb
@@ -2,7 +2,7 @@ Facter.add(:mongodb_version) do
   setcode do
     commands = ["rpmquery --qf='%{version}-%{release}' mongodb",
                 "repoquery --qf='%{version}-%{release}' mongodb",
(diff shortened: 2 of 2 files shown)


jlsherrill:
this seemed to solve the yum issue!  APT

ekohl:
Looks like tests fail because puppetlabs-concat should be pinned to 1.2.x

jlsherrill:
Note that the devel install still failed, but with the same error from http://ci.theforeman.org/view/Katello/job/systest_katello-devel/

iNecas:
FYI: The reason the custom facter was there is to be compatible both with upstream and downstream: the mongodb puppet module determines the paths form the os versions, which is not the case in downstream (where we don't take the mongodb from epel, as assumed in the puppet modules https://github.com/puppetlabs/puppetlabs-mongodb/blob/b41a706f7c1f3f01b84391972ecd6e135d113ab7/manifests/params.pp)

iNecas:
What was the error the devel installer was failing at?

ehelms:
I don't follow what the issue is between downstream vs up.

ehelms:
There is another way to fix the error, I was just hoping to get rid of work
arounds where possible.

iNecas:
In upstream, the mongodb version is given by what's available in the epel (in epel6 there is different version than epel7: 2.4 in el6 2.6 in el7 = differences in pidfile paths), while downstream we have a single version both in el6 and el7). The patch, that you are removing was done by the time the mongodb puppet module handled the different paths, but we were not able to use that (if we want to keep the modules as universal as possible)

ehelms:
This change isn't slated for downstream so if I am hearing that right, the
latest version of the puppet module (which we are using in nightly) handles
this correctly?

On Wed, May 6, 2015 at 10:08 AM, Ivan Necas <notifications@github.com>
wrote:

> In upstream, the mongodb version is given by what's available in the epel
> (in epel6 there is different version than epel7: 2.4 in el6 2.6 in el7 =
> differences in pidfile paths), while downstream we have a single version
> both in el6 and el7). The patch, that you are removing was done by the time
> the mongodb puppet module handled the different paths, but we were not able
> to use that (if we want to keep the modules as universal as possible)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Katello/puppet-pulp/pull/53#issuecomment-99478079>.
>



-- 
Go Pack and War Eagle,
Eric D. Helms

Ph.D. Student - North Carolina State University

iNecas:
No, it handles it correctly only if you're using epel as the source of mongodb (which we don't in downstream). The code you're removing should work both in upstream and downstream.

ehelms:
@iNecas updated with the other fix that will just squelch the yum errors that can occur so that the version is properly resolved

iNecas:
APT

PR #57, opened 2015-05-13
Fixing some documentation in admin.pp

timhughes:


ehelms:
ACK, thanks @timhughes

PR #58, opened 2015-05-18
Refs #10385: repoquery doesn't omit the errors, so we account for that.

ehelms:
Unlike the yum command, repoquery does not omit the errors that can
creep up. This change splits on newlines and grabs the last string
as that will be the version in all scenarios.


ehelms:
Not sure how I didn't see this the first time working on this change.

jlsherrill:
APJ

PR #59, opened 2015-05-23
Consumer

cristi1979:
create a new class pulp::agent that is needed by pulp consumer and/or pulp child node.


cristi1979:
that's not such a good idea after all

PR #61, opened 2015-05-23
updates for cosumer class

cristi1979:
- add validation for parameters
- let puppet manage the file /etc/yum/pluginconf.d/pulp-profile-update.conf
- install extra package for rabbitmq case


cristi1979:
can't squash, so i will do yet a new PR

PR #62, opened 2015-05-23
updates for consumer class

cristi1979:
- validate parameters
- install extra package for rabbitmq
- manage goferd service

### DIFF
--- a/manifests/consumer/service.pp
+++ b/manifests/consumer/service.pp
@@ -1,0 +1,9 @@
+# Pulp Consumer Service Packages
+class pulp::consumer::service {
+  service { 'goferd':
+    ensure     => running,
+    enable     => true,
+    hasstatus  => true,
+    hasrestart => true,
+  }
+}
--- a/manifests/consumer.pp
+++ b/manifests/consumer.pp
@@ -140,9 +140,16 @@ class pulp::consumer (
   $package_profile_enabled   = $pulp::consumer::params::package_profile_enabled,
   $package_profile_verbose   = $pulp::consumer::params::package_profile_verbose,
 ) inherits pulp::consumer::params {
-  include ::pulp::child::service
+  validate_bool($enable_puppet)
+  validate_bool($enable_nodes)
+  validate_bool($enable_rpm)
+
+  validate_bool($verify_ssl)
+  validate_bool($reboot_permit)
+  validate_bool($enable_color)
+  validate_bool($wrap_to_terminal)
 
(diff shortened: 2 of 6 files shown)


ehelms:
ACK, thanks @cristi1979

manifests/consumer/service.pp
```diff
@@ -0,0 +1,9 @@
+# Pulp Consumer Service Packages
+class pulp::consumer::service {
+  service { 'goferd':
```
ekohl: There is now a duplicate definition for the goferd service if both ```pulp::child::service``` and ```pulp::consumer::service``` are included.

cristi1979:
Can I suggest that pulp-katello and katello-agent could be part of the consumer class and could be enabled with a parameter $enable_katello.

pulp-nodes-child could be installed and configured from the main class along side with pulp server (which is required anyway) with something like $enable_child or a $parent_node parameter. In this case, the consumer class should be included and/or configured automatically with available data. Actually this was part of the patch I wanted to propose for making main class fully parametrized.

If not, I can use the "defined" function to install the package and start the service in both classes.

cristi1979:
My bad: pulp-katello should be also installed with pulp-server, not pulp consumer.

ekohl:
I don't really know pulp so I can't say what's the best design here.

ehelms:
I am hesitant to add mention of 'katello' to this module. Even though we maintain it, I prefer the modules be specific, but customizable, to the needs of it's users. If pulp-katello was named something a bit more generic, I would be ok including a way to install it with this module. Otherwise, I would prefer to err on the side of providing configurable options that another class could specify. For example, something like an `additional_packages` parameter that puppet-katello could pass (or some other wrapper class). If that is in the too complex to pull off, having a separate class inside this module for enabling 'katello' pieces would be best I think.

cristi1979:
Hi @ehelms, in case of katello-agent, the pulp-agent should not be installed (or can both os them exist on the same machine)? Should I do something like defining a parameter $agent_plugin (default to pulp) and install "${agent_plugin}-agent"? And pulp-consumer-client is still required for katello-agent?

ehelms:
I am not sure what would happen exactly. There is no 'conflicts' in the packaging of the katello-agent, but AFAIK the katello-agent is not designed to be added on top of pulp-agent. So, I would go with a parameter for allowing the passing in of the agent package name. I don't see `pulp-consumer-client` listed as a dependency for katello-agent.

PR #64, opened 2015-05-23
move httpd config to a separate class

cristi1979:
- use apache class to define the resources
- parametrize the configuration for rpm, docker, etc
- add a new parameter (ssl_verify_client) that can be used to allow access to https resources to clients that can't be verified:     ssl_verify_client     => 'optional_no_ca',


ehelms:
As we make these large changes, it would be nice to add even some basic tests to aide in hardening the module and to catch regressions going forward.

ehelms:
A manage_httpd or similar flag might be nice here. I know some Pulp setups just make use of the included configurations provided in the RPM.

ehelms:
Thoughts on using `concat::fragment` instead of a directory based approach?

cristi1979:
Sorry about the delay, I will try to finish the tests today.

Regarding the concat::fragment, should I use that for the files in  /etc/pulp/vhosts80/ and /etc/httpd/conf.d/pulp_* ?

cristi1979:
add flag httpd_manage
apache class rspec tests
include apache in init
add possibility to use different certificates for ssl (we are using the pulps autogenerated ca for users/consumers authorization and puppet certificates for ssl)

cristi1979:
Specs fail on ruby 1.9.3 because of this utf character: "# Copyright © 2010 Red Hat, Inc."

Should I remove it from the file?

ehelms:
+1 we can remove all the copyright notices from the tops of the templates

manifests/apache.pp
```diff
@@ -0,0 +1,154 @@
+# configure apache
+class pulp::apache {
+  include ::apache
+  include ::apache::mod::wsgi
+  include ::apache::mod::ssl
+
+  if $pulp::httpd_manage {
```
ehelms: Semantics but could we swap this around so it reads `manage_httpd` ?

manifests/apache.pp
```diff
@@ -0,0 +1,154 @@
...
+      }
+
+      file { '/etc/pulp/vhosts80/rpm.conf':
+        ensure  => file,
+        content => template('pulp/vhosts80/rpm.conf'),
+        owner   => 'apache',
+        group   => 'apache',
+        mode    => '0600',
+      }
+    }
```
ehelms: These plugin definitions end up looking very similar, maybe we should make them a `define` for re-use?

manifests/config.pp
```diff
@@ -1,6 +1,5 @@
 # Pulp Master Configuration
 class pulp::config {
-
   exec {'selinux_pulp_manage_puppet':
```
ehelms: Should this be guarded by the `enable_puppet` flag?

manifests/init.pp
```diff
@@ -26,6 +26,10 @@
 #
 # $consumers_ca_key::           The private key for the CA cert
 #
+#  $https_cert::                apache public certificate for ssl
+#
+#  $https_key::                 apache private certificate for ssl
```
ehelms: Nitpick of one less space before each of these

manifests/init.pp
```diff
@@ -70,6 +76,38 @@
...
+#
+#  $enable_child_node::         Boolean to enable pulp child nodes. Defaults
+#                               to false
+#                               type:boolean
+#
+#  $enable_http::               Boolean to enable http access to rpm repos. Defaults
+#                               to false
+#                               type:boolean
+#
+#  $httpd_manage::              Boolean to install and configure the httpd server. Defaults
```
ehelms: Similarly, an extra space exists before each of these

ehelms:
A few more comments, but I think this is looking really good.

cristi1979:
I squashed all commits

manifests/apache.pp
```diff
@@ -0,0 +1,100 @@
...
+          alias           => '/pulp/static',
+          path            => '/var/lib/pulp/static',
+          options         => ['Indexes'],
+          custom_fragment => 'SSLRequireSSL'
+        }
+      ],
+      options                    => ['SymLinksIfOwnerMatch'],
+      custom_fragment            => '# allow older yum clients to connect, see bz 647828
+	  SSLInsecureRenegotiation on
+	  AddDefaultCharset UTF-8',
```
ekohl: since puppetlabs-apache 1.2.0 you can set this on ```apache::vhost``` using ```add_default_charset```.

manifests/apache_plugin.pp
```diff
@@ -0,0 +1,21 @@
...
+      ensure  => file,
+      content => template("pulp/etc/httpd/conf.d/pulp_${name}.conf.erb"),
+      owner   => 'root',
+      group   => 'root',
+      mode    => '0644',
+    }
+  }
+
+  if $vhosts80 {
+    file { "/etc/pulp/vhosts80/${name}.conf":
```
ekohl: Any particular reason you're placing this in ```/etc/pulp```?

cristi1979: This is the default path where the installer puts the files.

ekohl: Ok

cristi1979:
Changed to use add_default_charset parameter

manifests/apache_plugin.pp
ehelms: I am assuming that instead of installer you meant where the Pulp RPMs put the files?

ehelms:
I am good with this PR after you squash commits please!

cristi1979:
Done

ehelms:
ACK, thanks @cristi1979 !

PR #65, opened 2015-05-23
Remove qpid and mongodb declaration

cristi1979:
I moved all db related declarations to a new class. This was done only to cleanup the config class. If it's not ok, I can revert this.

- force run of migrate_pulp_db after packages updates and configs updates
- use includes instead of declaring classes
-  add 3 new parameters: messaging_transport = 'qpid', broker_manage=true, db_manage=true (db and broker can be installed on another machine)

### DIFF
--- a/manifests/broker.pp
+++ b/manifests/broker.pp
@@ -1,0 +1,20 @@
+# Set up the broker
+class pulp::broker {
+  if $pulp::manage_broker {
+    if $pulp::messaging_transport == 'qpid' {
+      include ::qpid
+      $broker_service = 'qpidd'
+    } elsif $pulp::messaging_transport == 'rabbitmq' {
+      include ::rabbitmq
+      $broker_service = 'rabbitmq-server'
+    }
+
+    Service[$broker_service] -> Service['pulp_celerybeat']
+    Service[$broker_service] -> Service['pulp_workers']
+    Service[$broker_service] -> Service['pulp_resource_manager']
+  } else {
+    if $pulp::messaging_transport == 'qpid' {
+      package { 'qpid-tools': ensure => installed } -> Class['pulp::service']
+    }
+  }
+}
--- a/manifests/database.pp
+++ b/manifests/database.pp
@@ -1,0 +1,31 @@
+# Set up the pulp database
+class pulp::database {
+  if $pulp::manage_db {
+    if (versioncmp($::mongodb_version, '2.6.5') >= 0) {
+      $mongodb_pidfilepath = '/var/run/mongodb/mongod.pid'
+    } else {
+      $mongodb_pidfilepath = '/var/run/mongodb/mongodb.pid'
+    }
+
+    class { '::mongodb::server':
+      pidfilepath => $mongodb_pidfilepath,
+    }
+
+    Service['mongodb'] -> Service['pulp_celerybeat']
+    Service['mongodb'] -> Service['pulp_workers']
+    Service['mongodb'] -> Service['pulp_resource_manager']
+    Service['mongodb'] -> Exec['migrate_pulp_db']
+  }
+
+  exec { 'migrate_pulp_db':
+    command     => 'pulp-manage-db',
+    path        => '/bin:/usr/bin',
+    logoutput   => 'on_failure',
+    user        => 'apache',
+    refreshonly => true,
+    require     => File['/etc/pulp/server.conf'],
+  }
+
+  Class['pulp::install'] ~> Exec['migrate_pulp_db'] ~> Class['pulp::service']
+  Class['pulp::config'] ~> Exec['migrate_pulp_db'] ~> Class['pulp::service']
+}
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -13,6 +13,9 @@
 # $messaging_url::              URL for the AMQP server that Pulp will use to
 #                               communicate with nodes.
 #
+# $messaging_transport::        The type of broker you are connecting to. The default is 'qpid'.
+#                               For RabbitMQ, 'rabbitmq' should be used.
+#
 # $messaging_ca_cert:           The CA cert to authenicate against the AMQP server.
 #
 # $messaging_client_cert::      The client certificate signed by the CA cert
@@ -50,9 +53,6 @@
 # $repo_auth::                  Boolean to determine whether repos managed by
 #                               pulp will require authentication. Defaults
 #                               to true
-#
-# $reset_data::                 Boolean to reset the data in MongoDB. Defaults
-#                               to false
 #
 # $reset_cache::                Boolean to flush the cache. Defaults to false
 #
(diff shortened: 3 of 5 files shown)


cristi1979:
If this PR and also #64 are accepted I want to make a new patch that will make the pulp server fully configurable.

manifests/database.pp
```diff
@@ -0,0 +1,27 @@
...
+  exec { 'migrate_pulp_db':
+    command     => 'pulp-manage-db',
+    path        => '/bin:/usr/bin',
+    logoutput   => 'on_failure',
+    user        => 'apache',
+    refreshonly => true,
+    require     => File['/etc/pulp/server.conf'],
+  }
+
+  if $pulp::reset_data {
```
ehelms: You could remove this -- its legacy that we never removed when porting our original module.

manifests/init.pp
```diff
@@ -107,44 +119,45 @@
...
+  validate_bool($broker_manage)
 
   include ::apache
+  include ::mongodb::client
 
-  if (versioncmp($::mongodb_version, '2.6.5') >= 0) {
-    $mongodb_pidfilepath = '/var/run/mongodb/mongod.pid'
-  } else {
-    $mongodb_pidfilepath = '/var/run/mongodb/mongodb.pid'
+  if $db_manage {
```
ehelms: What about putting this block in the database class?

manifests/init.pp
```diff
@@ -107,44 +119,45 @@
...
+    if $messaging_transport == 'qpid' {
+      package { 'qpid-tools': ensure => installed } -> Class['pulp::service']
+    }
   }
-  class { '::apache::mod::wsgi':} ~>
-  class { '::mongodb':
-    logpath     => '/var/lib/mongodb/mongodb.log',
-    dbpath      => '/var/lib/mongodb',
-    pidfilepath => $mongodb_pidfilepath,
-  } ~>
```
ehelms: Unfortunately, we still need this logic due to the assumptions the mongodb module is making about what is available in upstream OS repositories.

manifests/init.pp
```diff
@@ -107,44 +119,45 @@
...
+
+    Service['mongodb'] -> Service['pulp_celerybeat']
+    Service['mongodb'] -> Service['pulp_workers']
+    Service['mongodb'] -> Service['pulp_resource_manager']
+    Service['mongodb'] -> Exec['migrate_pulp_db']
   }
 
-  class { '::mongodb::globals':
-    version => $::mongodb_version, # taken from the custom facts
+  if $broker_manage {
```
ehelms: Similar idea to the database, how about we move this to its own `broker` class?

ehelms:
I like this break out of concerns into areas and guarding them with manage flags.

cristi1979:
- removed $pulp::reset_data (should I do the same for reset_cache?)
- readded mongodb_version and call to mongodb class in order to preserve mongodb_pidfilepath. But I used mongodb::server instead of mongodb, which is deprecated.
- add a new class broker

manifests/database.pp
```diff
@@ -0,0 +1,31 @@
...
+  exec { 'migrate_pulp_db':
+    command     => 'pulp-manage-db',
+    path        => '/bin:/usr/bin',
+    logoutput   => 'on_failure',
+    user        => 'apache',
+    refreshonly => true,
+    require     => File['/etc/pulp/server.conf'],
+  }
+
+  Class['pulp::install'] ~> Exec['migrate_pulp_db'] ~> Class['pulp::service']
```
ekohl: I'd expect ```pulp::config``` to already require ```pulp::install``` so to me this line feels redundant.

manifests/init.pp
```diff
@@ -107,44 +115,20 @@
...
-    require => [Class['mongodb'], Class['qpid']],
-  } ~>
+  include ::pulp::broker
+  include ::pulp::database
+  class { '::pulp::install': } ->
+  class { '::pulp::config': } ~>
   class { '::pulp::service': } ~>
   Service['httpd']
-  ->
-  Class[pulp]
```
ekohl: This encloses the class and is in general a good thing to have.

manifests/database.pp
cristi1979: I don't understand what you mean.

if it's about Class['pulp::install'] and lass['pulp::config'], I wanted to force a migrate_pulp_db if any package is installed/updated or any configuration changes.

ekohl: Fair point about the package update. The main class only has a ```->``` relation, not ```~>```.

cristi1979:
I renamed manage flags in order to be consistent with the apache flag: manage_db and manage_broker

manifests/broker.pp
```diff
@@ -0,0 +1,20 @@
+# Set up the broker
+class pulp::broker {
+  if $pulp::manage_broker {
```
ehelms: Should we make this a parameter of this class? Same question for the 'messaging_transport' parameter? The reason I ask is this looks fairly independent and might could be used as an independent class to establish broker services? Granted, on the Katello side we do not currently do interesting Pulp setups at present outside of master and nodes.

cristi1979: It's not very clear for me what you are asking, but if you want to say to remove the manage_db and manage_broker parameters, let me know. 

From my point of view, pulp shouldn't manage the mongodb or broker installations (or any configuration aspect for that matter of those two).

ehelms: I am open to arguments as to why it should not; my personal take on it has always been that you should be able to configure things like brokers yourself, however, there should be an 'easy' button where Pulp configures everything it needs with sane defaults. In our current usage of the 0.1-stable branch, we rely on being able to do things like -- https://github.com/Katello/katello-installer/blob/master/modules/pulp/manifests/init.pp#L130 to customize to our needs.

cristi1979: Maybe the katello module should install and configure the dependencies.

The pulp module should only take care of pulp (again, that's from my point of view).

cristi1979: I think I took it too far with the above comment. A lot of puppet modules install those dependencies (foreman most notably) and the database should be manageable by pulp.

I was thinking about pidfilepath and logpath parameters that should not be managed by pulp. It should just install the db with default options and populate it.

ehelms: Since I don't have a true feel for this, I try to look at what other well defined modules do. For example, the puppetdb module from Puppetlabs (https://github.com/puppetlabs/puppetlabs-puppetdb) follows this type of pattern, where, because it requires a Postgres server to run will set one up unless a parameter is specified to indicate that the database shouldn't be managed by the module. I think that is where I tend to try to draw my line -- what is the minimum required to have a functioning Pulp and the module should set that up by default.

ehelms: I tend to agree. Everything looks good to me then, no need to worry about my prior comment. If you are good with this PR I'll merge it.

cristi1979: I don't have anything to add to this PR.

ehelms:
@cristi1979 looks like it needs a rebase since I merged the httpd PR

cristi1979:
I hope I did the rebase correctly.

manifests/init.pp
```diff
@@ -168,40 +178,17 @@
...
-    require => [Class['mongodb'], Class['qpid']],
-  } ~>
-  class { '::pulp::config':
-    require => [Class['mongodb'], Class['qpid']],
-  } ~>
+  validate_bool($manage_db)
+  validate_bool($manage_broker)
+
+  include ::apache
+  include ::mongodb::client
```
ehelms: We could argue that this should be moved to the service-wait puppet module.

cristi1979:
Any other issues with this PR?

ehelms:
Looks good! I'll merge it now, thanks!

PR #66, opened 2015-05-26
Refs #10616 - add ssl virt host fragment type

jlsherrill:


ekohl:
I have yet to look into this for puppet-foreman, but you could consider using concat fragments per https://github.com/puppetlabs/puppetlabs-apache#defined-type-apachevhost. This was not available in the puppetlabs-apache releases when I built the fragment define in puppet-foreman, but I think fragments is a cleaner solution.

ekohl:
Also, I now see there's a built in parameter for this type of config: ```additional_includes``` (which is rendered through https://github.com/puppetlabs/puppetlabs-apache/blob/master/templates/vhost/_additional_includes.erb)

jlsherrill:
@ekohl nice find on the additional_includes!  It looks like to have it use the 'IncludeOptional', i have to set the 'use_optional_includes' param on the apache class.  Is there a way to do that without declaring the top level apache class?

https://github.com/puppetlabs/puppetlabs-apache/blob/1962d7309a040878ccc73ccadc47a28c6caaceaa/manifests/init.pp#L74


I think i prefer this fragment 'type' over the concat fragments for one reason, the place where you add a fragment, you need to know the vhost name and defined order number, if either of those changes your fragment breaks.  Sure tests can catch that, but this seems cleaner.

ehelms:
You could create a parameter that represents the configuration file and then use that:

```
pulp::child (
   ..........
) { 
....
$vhost_config = "${priorty}-${vhost_name}"
}
```

And then in another module:

```
concat::fragment { 'gpg_key_config':
  target => pulp::child::vhost_config,
  content => 'GPG KEYS'
}
```

ekohl:
As @ehelms says, I think ```pulp::child::fragment``` should wrap ```concat::fragment``` with the proper target.

jlsherrill:
@ekohl i do like that idea of pulp::child::fragment wrapping  concat::fragment, and i've implemented that here.  Wasn't sure about the 'order', so i made it a param with a default value.

manifests/child/fragment.pp
```diff
@@ -0,0 +1,19 @@
...
+) {
+
+  concat::fragment { $name:
+    target  => '25-pulp-node-ssl.conf',
+    content => $ssl_content,
+    order   => $order,
+  }
+
+}
+
```
ekohl: Nitpick: empty line

spec/defines/pulp_child_fragment_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+        { :ssl_content => "some_string"}
+      end
+
+      it { should contain_concat__fragment("test_name").
+             with_content(/some_string/) }
+    end
+
+  end
+end
+
```
ekohl: Nitpick: empty line

ekohl:
@jlsherrill I think it looks good, but it'd be good to test this well and check if the proper minimal puppetlabs-apache version is in metadata.json.

jlsherrill:
bumped the minimum apache module version to match when that feature was first added.  Also tested doing multiple fragments and it seemed to work fine.

ekohl:
:+1: looks good. Nice and simple.

spec/defines/pulp_child_fragment_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+        :operatingsystemmajrelease => '6.4',
+        :osreleasemajor            => '6',
+        :osfamily                  => 'RedHat',
+        :processorcount            => 3,
+      }
+    end
+ 
+    context 'with ssl_content parameter' do 
+      let :params do 
+        { :ssl_content => "some_string"}
```
ehelms: Nitpick - even spacing for brackets.

spec/defines/pulp_child_fragment_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+      end
+
+      it { should contain_concat__fragment("test_name").
+             with_content(/some_string/) }
```
ehelms: Nitpick - multiline block use `do-end`

ehelms:
ACK pending the two minor comments -- this will need to be cherry picked back to 0.1-stable

jlsherrill:
updated, will open new PR for backport

PR #69, opened 2015-06-16
refs #10716 - migrate after plugin install

stbenjam:


stbenjam:
@ehelms So, this is kind of what I was thinking for how to handle things for the migrations when plugins are installed, but maybe we should wait for #65 to be ready, so we can use the new structure of the module and notify the existing migrate DB (can't notify the current one because of the `creates` limitation).  It'd be cleaner.

But, we need to solve the problem for downstream, do you think we could take https://github.com/Katello/katello-installer/pull/233 now and do this pulp stuff for 2.3?

ehelms:
I think we'd want #64 as it adds per plugin enablement params (https://github.com/Katello/puppet-pulp/pull/64/files#diff-60ae41fd0a31977447947f59940ee9a4R143). However, I am not sure how much time @cristi1979 gets to work on these.

I am OK with the installer PR in light of this as long as we keep an issue open that we are tracking actively.

manifests/plugin.pp
```diff
@@ -0,0 +1,26 @@
...
+
+  $real_package = $package ? {
+      undef       => "pulp-${name}-plugins",
+      default     => $package,
+  }
+
+  package { $real_package:
+    ensure => installed,
+  } ~>
+  exec {'migrate_pulp_db':
```
ekohl: Doesn't this result in duplicate resources with multiple plugins?

manifests/plugin.pp
```diff
@@ -0,0 +1,26 @@
...
+    command     => 'pulp-manage-db && touch /var/lib/pulp/init.flag',
+    path        => '/bin:/usr/bin',
+    logoutput   => 'on_failure',
+    user        => 'apache',
+    require     => [Service[mongodb], Service[qpidd], File['/etc/pulp/server.conf']],
```
ekohl: I believe ```mongodb``` and ```qpidd``` should be quoted.

manifests/plugin.pp
```diff
@@ -0,0 +1,26 @@
+#
+# == Define: pulp::plugin
+#
+# Install and configure a Pulp plugin
+#
+define pulp::plugin (
+  $package  = undef,
```
ekohl: You can simplify this to ```$package = "pulp-${title}-plugins"```

manifests/plugin.pp
stbenjam: Yes, sorry, this wasn't meant for review.  See my comment to @ehelms, #65 removes the `creates` restriction on the other resources, so we can just notify it once its merged.  #64 and #65 need to go first and then I'll refactor this.

ehelms:
@stbenjam I think we the recent merges you can re-visit this change now.

cristi1979:
I think this is already taken care of with this code from database.pp?

    Class['pulp::install'] ~> Exec['migrate_pulp_db'] ~> Class['pulp::service']
    Class['pulp::config'] ~> Exec['migrate_pulp_db'] ~> Class['pulp::service']

ehelms:
@stbenjam is this still needed given the changes to master?

stbenjam:
Nope, thanks!

PR #70, opened 2015-06-20
new pulp-admin rpm repo provider

cristifalcas:


cristi1979:
I just noticed that I left a parameter conf_file. Nothing was implemented for this. Do you think this is useful? Also, I don't think that purging will work correctly with different config files.

cristi1979:
If there is no login token, prefetching will not work. So, if the an existing repo is created, it will throw an error.

cristi1979:
Usage:

    pulp_rpmrepo { 'pulp_el6':
      ensure           => 'present',
      checksum_type    => 'sha256',
      feed             => 'http://repos.fedorapeople.org/repos/pulp/pulp/stable/latest/6Server/x86_64/',
      generate_sqlite  => 'false',
      max_downloads    => '10',
      relative_url     => 'pulp/6Server',
      remove_missing   => 'true',
      retain_old_count => '1',
      serve_http       => 'true',
      serve_https      => 'true',
      validate         => 'true',
    }

ehelms:
I have only taken a cursory glance at this, but I have a question before I proceed further. The provider here and I will assume future providers are writing a good chunk of Ruby code to talk to the Pulp API and handle the configurations therein. We have a and actively maintain a set of Ruby bindings to talk to the Pulp API. Would it be better to just make use of that library and centralize that aspect? https://github.com/Katello/runcible

cristi1979:
I don't see how the user will install that.

ehelms:
The puppet-foreman_proxy has a similarish type of concern where a provider makes use of an API gem which they include via a package directive -- https://github.com/theforeman/puppet-foreman_proxy/blob/master/manifests/install.pp#L20

cristi1979:
Is it bundled in pulp repos? I can change the providers and the module to use that gem.

ehelms:
It is not currently, but I could chat with them about carrying it.

cristi1979:
OK, I will wait for your update.

ehelms:
@cristi1979 can you clarify for me so I am understanding this right, in some cases its using pulp-admin to do tasks and some the API directly?

cristi1979:
Yes, but only because I'm lazy. The final command, that id executed with pulp-admin. I will update the provider to use the api also.

ehelms:
The reason I ask the question is because if everything can be boiled down to use pulp-admin, then we don't need to ship Runcible in the pulp repos. If the API makes more sense to use directly instead, then I will proceed with getting Runcible into their repositories.

cristi1979:
The output from pulp-admin is useless in determining programmatically the current resource properties. The api calls are mandatory for reading them.

PR #71, opened 2015-06-20
add a simple login/logout provider

cristifalcas:
a simple provider to login/logout a pulp user

### DIFF
--- a/lib/puppet/provider/pulp_login/pulp_admin.rb
+++ b/lib/puppet/provider/pulp_login/pulp_admin.rb
@@ -1,17 +1,17 @@
+require 'pp'
 Puppet::Type.type(:pulp_login).provide(:pulp_admin) do
   desc "Login/logout a pulp user"
 
   commands :pulp_admin => '/bin/pulp-admin'
   def exists?
-    begin
-      pulp_admin(['repo', 'list'])
-    rescue => err
-      false
-    end
-    true
+    # always force relogin with the proper user
+    pulp_admin('logout')
+    false
   end
 
   def create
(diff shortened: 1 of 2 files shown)


cristi1979:
This is useless. I can't figure it out how to determine if the correct user has already logged in.

cristi1979:
This was changed to always login the user: exists? returns always false.

cristi1979:
A side effect of this is that logout will never be triggered. But, since I also force a logout in exists...

It doesn't look like a proper provider. Maybe an exec will be more useful instead?

cristi1979:
Anyway, this is a sample usage:

	  pulp_login { 'admin': pass => 'admin' } ->
	  pulp_rpmrepo { 'epel':
	    ensure           => 'present',
	    validate         => true,
	    remove_missing   => true,
	    serve_http       => true,
	    retain_old_count => 5,
	    relative_url     => "epel/6Server",
	    feed             => 'http://dl.fedoraproject.org/pub/epel/6/x86_64/',
	  } ->
	  pulp_login { 'admin_logout': ensure => absent }

ekohl:
I'm not sure how useful this is since I don't know pulp well enough, but shouldn't ```pulp_rpmrepo``` handle the authentication? I know this is hard problem (since we have a similar problem with foreman types through the API), but this solution feels like imperative programming in a declarative language.

cristifalcas:
It's not useful at all, since it's only a hidden exec.

ekohl:
Should we close this? Refactor this into something more useful?

cristifalcas:
I will close it

PR #74, opened 2015-06-24
pulp sync schedule provider

cristifalcas:
This is an abomination because of some limiting factors: 
- I can't make the connection between repo id + schedule to a resource name
- I don't think that multiple schedules per repo can be implemented (from my point of view it can't be done)

Limitations:
- we can't have multiple schedules for the same repo, so we force the existence of only one schedule per repo (the one with the smallest sync_id). The resource name will be the repo_id
- user should not set the name of a repo to be the same as an existing schedule_id (probability is very small)
- if the repo (resource name) doesn't exists, the resource will fail (normal)
- there is a global variable used ($schedules_info)

Explanation on how it's implemented:
- when we get the resources, we set the resource name to repo_id only for the first schedule
- the other schedules get the name = schedule_id (this is useful only for purging)
- we need to keep somewhere the schedule_id for the first ones ($schedules_info['schedule_id']): this is needed in update
- we need to keep the repo_id for the second ones ($schedules_info['repo_id']): needed in delete, when we purge the resources
- since this provider is global for all repos, we need to keep the repo_type for all repos ($schedules_info['repo_type'])
- a special case is when we have a single resource defined in puppet with ensure => absent: we keep the schedule id in $schedules_info['repo_id']

Example usage:

```
  pulp_schedule { 'katello_client_el6':
    schedule_time => '2002-01-01T00:00Z/P1D',
    ensure        => present
  }

  pulp_schedule { 'katello_client_el7':
    ensure        => absent
  }
```

### DIFF
--- a/lib/puppet/provider/pulp_schedule/api.rb
+++ b/lib/puppet/provider/pulp_schedule/api.rb
@@ -1,0 +1,164 @@
+#  Limitations:
+# - we can't have multiple schedules for the same repo, so we force the existence of only one schedule per repo 
+#   (the one with the smallest sync_id). The resource name will be the repo_id
+# - user should not set the name of a repo to be the same as an existing schedule_id (probability is very small)
+# - if the repo (resource name) doesn't exists, the resource will fail (normal)
+# - there is a global variable used ($schedules_info)
+#
+# Explanation on how it's implemented:
+# - when we get the resources, we set the resource name to repo_id only for the first schedule
+# - the other schedules get the name = schedule_id (this is useful only for purging)
+# - we need to keep somewhere the schedule_id for the first ones ($schedules_info['schedule_id']): this is needed in update
+# - we need to keep the repo_id for the second ones ($schedules_info['repo_id']): needed in delete, when we purge the resources
+# - since this provider is global for all repos, we need to keep the repo_type for all repos ($schedules_info['repo_type'])
+# - a special case is when we have a single resource defined in puppet with ensure => absent: we keep the schedule id in $schedules_info['repo_id']
+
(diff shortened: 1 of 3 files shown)


cristi1979:
I pushed a new version that takes care of schedules for any repo type (rpm, puppet, etc):
      pulp_schedule { 'katello_client_el6':
	schedule_time => '2002-01-01T00:00Z/P1D',
	ensure        => present
      }

      pulp_schedule { 'katello_client_el7': 	ensure        => absent       }

This is with an extra api call that gets the repo type.

cristi1979:
There is still a problem when purging repos and schedules: if the repo is deleted first, the schedule will fail to do the purging. I think this is because when deleting the repo, pulp cleans the schedules also.

I don't know how to fix this yet. I could to an extra check in delete to see if the schedule still exists, but it looks ugly. Also I don't like to add more special cases then there are already.

ehelms:
@cristifalcas any plans to re-visit the issues you mentioned with this? Or is it mergeable as is for now?

ccnifo:
This PR is very interesting from my point of view, but can it be useful without a Puppet managed repo creation, like in the closed PR #70 ?

ehelms:
I'm willing to merge this if someone is willing to vet it still works.

cristifalcas:
We are using those providers on our company on more then 10 pulp servers. Let me check if I done other modifications on our side and merge those too.

ehelms:
@cristifalcas great! I know we had a lot of discussion around them previously, however, if you have found success with them and others want them (which it seems they do) then I'm up for merging them into this module.

PR #79, opened 2015-07-03
fix including apache configuration

cristi1979:
fix including apache configuration


ehelms:
ACK, thanks @cristi1979

PR #81, opened 2015-07-03
install extra packages in case of qpid broker

cristi1979:


ehelms:
I am missing where `pulp::consumer` gets called by the main pulp init.pp class.

cristi1979:
Nowhere. I want to make a new PR with a new parameter (enable_nodes) that will include the consumer class.

Consumer class is not required on a normal pulp server.

ehelms:
Will that fix the issue I am seeing where with the parameters PR I can't do a basic puppet apply of the pulp class? e.g. `puppet apply -e "class {pulp: }" ?  Currently that fails because of this package not being installed.

cristi1979:
Sorry, I misunderstood what you said. I thought that goferd throws errors because of the missing package.

PR #82, opened 2015-07-05
Require qpidd to be started before migrating the database.

ehelms:


cristi1979:
I think that this belongs in broker.pp, because we don't know the broker type.

ehelms:
@cristi1979 good call -- updated to move it into the broker

cristi1979:
We also need to switch the order of the broker and database includes in init.pp

ehelms:
Good catch, tests for the win! Updated, should pass this go around.

cristi1979:
And it should be Exec[migrate_pulp_db]

ehelms:
Damn. I should not be allowed to edit puppet modules before 11 am.

cristi1979:
It looks good for me.

ehelms:
Thanks for the review @cristi1979

PR #83, opened 2015-07-07
Consumer

cristi-falcas:
install extra packages in case of qpid and rabbitmq broker

### DIFF
--- a/manifests/broker.pp
+++ b/manifests/broker.pp
@@ -12,6 +12,7 @@ class pulp::broker {
     Service[$broker_service] -> Service['pulp_celerybeat']
     Service[$broker_service] -> Service['pulp_workers']
     Service[$broker_service] -> Service['pulp_resource_manager']
+    Service[$broker_service] -> Exec['migrate_pulp_db']
   } else {
     if $pulp::messaging_transport == 'qpid' {
       package { 'qpid-tools': ensure => installed } -> Class['pulp::service']
--- a/manifests/consumer/install.pp
+++ b/manifests/consumer/install.pp
@@ -1,11 +1,19 @@
 # Pulp Consumer Install Packages
 class pulp::consumer::install {
-  if $pulp::consumer::messaging_transport == 'rabbitmq' {
-    package { 'python-gofer-amqp':
-      ensure => $pulp::consumer::version,
+  if $pulp::consumer::messaging_transport == 'qpid' {
+    ensure_packages(['python-gofer-qpid'], {
+      ensure => $pulp::consumer::version
     }
+    )
   }
 
+  if $pulp::consumer::messaging_transport == 'rabbitmq' {
+    ensure_packages(['python-gofer-amqp'], {
+      ensure => $pulp::consumer::version
+    }
+    )
+  }
+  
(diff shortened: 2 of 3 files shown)


ehelms:
@cristi1979 In case you were curious, this what I am using to test PRs and vet our Ruby API wrapper currently -- https://github.com/Katello/katello-deploy/pull/165

cristi1979:
After the bugs you found after running the code I started using a docker image where I run the pulp module and also our pulp profile (with rabbitmq+ssl and mongodb+ssl).

Should I understand that pulp::consumer class is not working anymore? :) 

I will do a test now also.

ehelms:
Oh, I didn't mean to imply that, I just picked a random PR to comment to you about it :)

cristi1979:
With the latest commit it should work. I tested register, bind and unregister.

ehelms:
I would be interested in your container based testing strategy if you are able to share.

cristi1979:
I have an el7 container that is build with the info from here:
http://developerblog.redhat.com/2014/05/05/running-systemd-within-docker-container/

For tests I'm using the script from https://gist.github.com/cristi1979/e15d2a9ba36aeeb11e0b. Currently I run the commands manually. Currently I do the tests on el7 only, but I will start using an el6 container also.

What is the script doing:
- I mount the folder where I keep the puppet modules in /mnt when I start the container
- there is a custom script that enables epel and pulp repos
- since we're using the puppet certificates for all our ssl communications, I also create those
- run the pulp classes
- do some pulp-admin and pulp-consumer tests

An example of how I defined the pulp class is here: https://gist.github.com/cristi1979/90d7d651814418d28694

We have an extra class that copies the pulp certificates to /etc/pki/puppet_certs/$user and gives only user $user full rights on those files

Let me know if you have any questions.

cristi1979:
There are some extra parameters there (enable_parent_node, enable_child_node) for which I want to make a PR latter.

cristi1979:
Maybe it depends on the base image you choose, but for puppet I also had to do this in our image:

    RUN yum reinstall glibc-common -y
    RUN (cd /usr/share/locale/; for i in *; do [ $i == en_US ] || rm -rf $i; done) && \
             echo export LC_ALL=en_US.utf8 >> ~/.bashrc

cristi1979:
I know I'm spaming now, but I decided to "automate" those steps and this is what I came up with:

I have the script from https://gist.github.com/cristi1979/b3effd1661a260b43baa in $HOME/test_pulp.sh. And I run those 3 commands:

    ID=$(docker run --privileged -d=true -v /sys/fs/cgroup:/sys/fs/cgroup:ro -v /home/puppet_environments/falcas/:/mnt/:ro -v ~/test_pulp.sh:/test_pulp.sh v-so-repo-06.synygy.net/baseimages/el7builder:latest)
    docker exec $ID /test_pulp.sh
    docker rm -f $ID

cristi-falcas:
Any problems with this one?

ehelms:
Looks good, squash and I'll merge

cristi-falcas:
done

ehelms:
ACK, thanks @cristi-falcas

PR #85, opened 2015-07-08
Config file updates, ostree support, and bugfix

beav:
The repo_auth conf file has been updated to match what is in Pulp 2.7. The
vhost configuration also now allows for adding custom third party wsgi scripts
via the additional_wsgi_scripts hash.

Additionally, changes were made to support ostree and a small bugfix was made
related to ssl_ca_cert.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -107,6 +107,12 @@
 #
 # $consumer_cert_expiration::   number of days a consumer certificate is valid
 #
+# $disabled_authenticators::    List of repo authenticators to disable.
+#                               type:array
+#
+# $additional_wsgi_scripts::    Hash of additional paths and WSGI script locations for Pulp vhost
+#                               type:hash
+#
 # $reset_cache::                Boolean to flush the cache. Defaults to false
 #                               type:boolean
 #
@@ -162,6 +168,10 @@
 #                               type:boolean
 #
 # $enable_python::              Boolean to enable python plugin. Defaults
+#                               to false
+#                               type:boolean
+#
+# $enable_ostree::              Boolean to enable ostree plugin. Defaults
 #                               to false
 #                               type:boolean
 #
@@ -259,6 +269,7 @@ class pulp (
   $enable_rpm                = $pulp::params::enable_rpm,
   $enable_puppet             = $pulp::params::enable_puppet,
   $enable_python             = $pulp::params::enable_python,
+  $enable_ostree             = $pulp::params::enable_ostree,
   $enable_parent_node        = $pulp::params::enable_parent_node,
   $enable_http               = $pulp::params::enable_http,
   $manage_broker             = $pulp::params::manage_broker,
@@ -271,19 +282,24 @@ class pulp (
   $node_oauth_effective_user = $pulp::params::node_oauth_effective_user,
   $node_oauth_key            = $pulp::params::node_oauth_key,
   $node_oauth_secret         = $pulp::params::node_oauth_secret,
+  $disabled_authenticators   = $pulp::params::disabled_authenticators,
+  $additional_wsgi_scripts   = $pulp::params::additional_wsgi_scripts,
 ) inherits pulp::params {
-  validate_bool($repo_auth)
-  validate_bool($reset_cache)
   validate_bool($enable_docker)
   validate_bool($enable_rpm)
   validate_bool($enable_puppet)
   validate_bool($enable_python)
+  validate_bool($enable_ostree)
   validate_bool($enable_http)
   validate_bool($manage_db)
   validate_bool($manage_broker)
   validate_bool($manage_httpd)
   validate_bool($manage_plugins_httpd)
   validate_bool($enable_parent_node)
+  validate_bool($repo_auth)
+  validate_bool($reset_cache)
+  validate_array($disabled_authenticators)
+  validate_hash($additional_wsgi_scripts)
 
   include ::mongodb::client
   include ::pulp::apache
--- a/templates/repo_auth.conf.erb
+++ b/templates/repo_auth.conf.erb
@@ -1,3 +1,5 @@
+# This file is managed by puppet. Changes may be overridden!
+
 [main]
 enabled: <%= @repo_auth ? "true" : "false" %>
 log_failed_cert: true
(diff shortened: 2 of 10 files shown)


manifests/init.pp
```diff
@@ -271,18 +280,21 @@
   $node_oauth_effective_user = $pulp::params::node_oauth_effective_user,
   $node_oauth_key            = $pulp::params::node_oauth_key,
   $node_oauth_secret         = $pulp::params::node_oauth_secret,
-) inherits pulp::params {
-  validate_bool($repo_auth)
-  validate_bool($reset_cache)
+  $disabled_authenticators   = $pulp::params::disabled_authenticators,
+  $additional_wsgi_scripts   = $pulp::params::additional_wsgi_scripts,
+  ) inherits pulp::params {
```
ekohl: Nitpick: odd indenting here

manifests/init.pp
```diff
@@ -271,18 +280,21 @@
   $node_oauth_effective_user = $pulp::params::node_oauth_effective_user,
   $node_oauth_key            = $pulp::params::node_oauth_key,
   $node_oauth_secret         = $pulp::params::node_oauth_secret,
-) inherits pulp::params {
-  validate_bool($repo_auth)
-  validate_bool($reset_cache)
+  $disabled_authenticators   = $pulp::params::disabled_authenticators,
+  $additional_wsgi_scripts   = $pulp::params::additional_wsgi_scripts,
```
ekohl: Validate both with ```validate_hash```?

beav:
Is anything else needed for this PR?

ekohl:
I always forget that github doesn't notify you on new commits.

manifests/init.pp
```diff
@@ -107,6 +107,10 @@
 #
 # $consumer_cert_expiration::   number of days a consumer certificate is valid
 #
+# $disabled_authenticators::    List of repo authenticators to disable.
+#
+# $additional_wsgi_scripts::    Hash of additional paths and WSGI script locations for Pulp vhost
```
ekohl: This could use a ```type:hash```

manifests/init.pp
```diff
@@ -107,6 +107,10 @@
 #
 # $consumer_cert_expiration::   number of days a consumer certificate is valid
 #
+# $disabled_authenticators::    List of repo authenticators to disable.
```
ekohl: This could use a ```type:array```

ekohl:
Minor nitpicks, otherwise the code looks good.

beav:
@ekohl updated, thanks for the review :koala:

ekohl:
:+1:

ehelms:
The additions made herein mean that master of this module would "only" be usable for 2.7? In other words, there are some 2.7 specific items that are not gated based on the Pulp version available?

beav:
@ehelms  The 2.7 additions are not gated based on version, but must be enabled specifically. For example, ostree support is disabled by default, and the additional config directive in repo_auth.conf is not included by default.

I have not tested this module with Pulp 2.6, but I believe the changes in this PR would not cause any incompatibilities.

ehelms:
So maybe we should gate them? I know other modules that set parameters or do validations based on version of the underlying service being setup because of things like this (e.g. prevent users form enabling os tree if doing a 2.6 setup)

beav:
@ehelms we can do that. I will make the appropriate updates.

ehelms:
@beav planning to re-visit this still?

beav:
@ehelms I am but I need to wrap up some unrelated items first. I think I will be able to revisit this in two weeks. Does that work ok?

ehelms:
Sounds good! Thanks for the update

ehelms:
@beav looks like this needs a rebase and to address @ekohl comment above -- after that I think it should be good

ekohl:
@ehelms github doesn't notice there is an extra line, but it looks like it was already there.

ehelms:
@ekohl duh on me, good catch, thanks for pointing that out :)

templates/repo_auth.conf.erb
```diff
@@ -11,6 +13,13 @@ max_num_certs_in_chain: 100
 # maintain backwards compatibility.
 verify_ssl: false
 
+# If set, this disables specific repo auth plugins. More than one plugin can be
+# specified in the form of "plugin1,plugin2,plugin3".
+# disabled_authenticators = oid_validation
+<% if @disabled_authenticators and ! @disabled_authenticators.empty? -%>
```
ehelms: Since ERB is just ruby, I think we want to avoid 'and' keyword usage in favor of '&&'.

ehelms:
Is the repo_auth file change for 2.7 that you mentioned only available in 2.7? If so, and we sorta talked about this before I think we should introduce either:

a) versioning within the puppet module
b) include in the README which version of Pulp this module works for

stbenjam:
I'd vote for versioning of the module if the change is not backwards compatible.  E.g., puppet-pulp 1.x -> pulp 2.6, puppet-pulp 2.x -> pulp 2.7.  

Having puppet be intelligent enough to know what version of software its installing never works out well since the catalog can be compiled on the master.

beav:
@stbenjam I'm ok with either way but once 2.7 comes out (which should be soon-ish), 2.6 is no longer updated except for severe bugs.

I think the option with the least amount of work would be to wait for 2.7 GA to be released and then merge this. I have no problem with versioning and keeping the 2.6.x around, but IMO I dont want to encourage people to stay on 2.6 by keeping a 2.6 branch after 2.7 is out.

ehelms:
We don't have to keep a branch unless we want to maintain puppet-pulp 1.X with backports; however, even if we wait for 2.7 GA, I think I agree with @stbenjam that we should bump puppet-pulp to 2.X since it would include incompatible changes.

beav:
Should this PR bump metadata.json to 2.0.0?

ehelms:
I'll let @stbenjam answer that as I am still figuring out the full workflow with the use of blacksmith.

stbenjam:
I would probably bump it here.

That means next time we do module:release with blacksmith it'd publish 2.0.0 to the forge and not accidentally release 1.0.something.

You *could* do it with blacksmith but module:bump_major isn't doing anything fancy.

beav:
Added an update to metadata.json and a "DO NOT MERGE" name until 2.7.0 is GA.

ehelms:
Looks like this needs to be rebased

beav:
updated for rebase

beav:
also removed 'do not merge' since 2.7.0 is out

ehelms:
@beav has Pulp 2.7 been released?

beav:
@ehelms it has indeed :panda_face: 

http://www.pulpproject.org/

ehelms:
ACK - thanks @beav !

PR #86, opened 2015-07-16
fixes #10716 - migrate pulp when new pulp plugins are installed

stbenjam:
This is a better alternative to https://github.com/Katello/katello-installer/pull/233.

With https://github.com/Katello/katello-installer/pull/233, we still encounter errors in the logs because the pulp service is started by the installer before the post hook runs.  The post hook fixes it, and pulp is working fine, but the errors in the logs are very ugly.

This is only for pulp-puppet 0.1.  The new version 1.0 does the right thing when additional packages are installed.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,6 +7,7 @@ group :test do
   gem "rspec-puppet", :git => 'https://github.com/rodjek/rspec-puppet.git'
   gem "puppet-syntax"
   gem "puppetlabs_spec_helper"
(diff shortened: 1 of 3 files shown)


stbenjam:
cherry-picked 44b2c9269982c94822f94ec18930c09fe89a1f5f to make tests pass on  1.8.7

ehelms:
ACK

PR #87, opened 2015-07-16
fix config.pp

cristi-falcas:
nothing in config.pp was added, only shifted.


ekohl:
Looks good :+1:

ehelms:
ACK, thanks @cristi-falcas !

Issue #91, opened 2015-08-07
Machines without the pulp class included still run queries for mongodb version


andymillar:
pulp registers a fact to identify the mongodb version so that it can configure $mongodb_pidfilepath in pulp::database

As this is defined as a fact it is run on every server associated with the puppetmaster, not just those including the pulp class.

This puts unnecessary load on systems and upstream repo servers as yum is hit with every facter run.

ehelms:
@andymillar feel free to open a PR with your change and we'll review it! @iNecas thoughts?

andymillar:
@ehelms done. Took a little longer to commit and make sure it actually worked.

andymillar:
Converting the fact to a function doesn't work as desired as it gets run on the puppetmaster and not the client node. 

I've closed the PR, going to have to find a different way of doing this.

ekohl:
@andymillar I wonder if you could somehow cache the data and easily detect freshness.

stbenjam:
yum is not called on every run, it tries rpmquery first, which puts no load on anything but the local host, and a very marginal amount. 

yum also caches metadata for some time, so it handles the caching of the metadata for us.  Is this causing a problem for you?

If you've got 10,000 hosts refreshing their yum metadata a few times a day, that is a lot of traffic.  I'd just remove the yum command, honestly. I'm not sure why the yum as a last resort is there, because I'd think they'd all have rpmquery if they have yum.

It's puppet's design that all facts go everywhere without much way to control it.  You can add a `:confine` to something like a Red Hat OS, which could solve some cases to avoid running these commands on an Ubuntu box, but otherwise I don't think you can do anything about this.

stbenjam:
The amount of headaches this stupid pidfile location change has caused... :angry:  Does it make mongo somehow more webscale when you just break everyone?

We could just strip out all the logic here - because puppetlabs/puppetlabs-mongodb seems to **partially** handle the new improved mongo pidfile.  The problem is that it does assume all fedora and > EL7 operating systems use the new version.

RHEL does not, in Satellite.  If we can get them to rebase on the newer mongo, then the > EL 7 assumption would be (more) correct and this could go away.

andymillar:
Better question, actually, might be why do we care what the mongodb pidfile is?

The pidfile passed to the puppetlabs-mongodb class is used to write the mongodb config file. 

[...]

PR #92, opened 2015-08-07
Replace mongodb_version fact with get_mongodb_version function to prevent lookups of the mongodb version on hosts that do not include the pulp class. Fixes GH-91

andymillar:
As issue-91, replace the mongodb_version fact with an equivalent get_mongodb_version.

This prevents hosts not running pulp from querying their available mongodb versions. 

This also removes the non-obvious "facts.rb" fact file name, cleaning up host facts directory. 

Travis tests appear to pass, and the following basic test also works:

modules$ cat pulp/manifests/foo.pp 
class pulp::foo {
  $foo = get_mongodb_version()
  notify { $foo: }
}

modules$ puppet apply --modulepath=. -e "include pulp::foo" --noop
Notice: /Stage[main]/Pulp::Foo/Notify[2.4.14-1.el6]/message: current_value absent, should be 2.4.14-1.el6 (noop)
Notice: Class[Pulp::Foo]: Would have triggered 'refresh' from 1 events
Notice: Stage[main]: Would have triggered 'refresh' from 1 events
Notice: Finished catalog run in 0.06 seconds

### DIFF
--- a/lib/puppet/parser/functions/get_mongodb_version.rb
+++ b/lib/puppet/parser/functions/get_mongodb_version.rb
@@ -1,5 +1,5 @@
-Facter.add(:mongodb_version) do
-  setcode do
+module Puppet::Parser::Functions
+  newfunction(:get_mongodb_version, :type => :rvalue) do |args|
     commands = ["rpmquery --qf='%{version}-%{release}' mongodb",
                 "repoquery --qf='%{version}-%{release}' mongodb",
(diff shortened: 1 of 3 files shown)


andymillar:
This doesn't actually do what we want.

Functions are executed by the compiler — in an agent/master Puppet environment, this means they run on the puppet master server, and can’t access files or resources on any agent node. Functions can only interact with agent-provided information in the form of facts.

PR #93, opened 2015-08-10
Remove mongodb_version fact as it's not necessary. The mongodb class handles this for us

andymillar:
Fixes GH-91

The mongodb class actually cares about the version, and guesses that itself. 

Passing the pidfile location only sets a template variable (so it really doesn't matter what is set).

We may as well not set it, and let them work it out for us. They do for everything else :-)

Andy


stbenjam:
Sorry NACK from me, please re-read my comment.  This is absolutely required, because we need to know the mongo version to set the pidfile correct.  This isn't sufficient.

stbenjam:
Didn't mean to hit close, I'll let the others review and share their opinion.

We can do this if we can make sure puppetlabs-mongodb operating system-version based check is correct, which it currently is not for RHEL.

stbenjam:
This has sat for a while without any feedback from anyone else, so I'm gonna close the PR for now.  I'll leave the issue open, as it would be fantastic for us to someday not need to carry the version fact.  

Unfortunately as discussed in GH-91, the limiting factor is still the PL module's assumption of a mongo version based on the OS version which isn't true in some of the cases this module has got to support.

Thanks @andymillar

PR #96, opened 2015-09-04
add the posibility to install katello_agent

cristifalcas:


ehelms:
Does this use anything from the base pulp module? Seems like it might make more sense broken out to it's own module? More than happy to create and host that on our github organization if we decide that route is the best way to go.

cristifalcas:
nothing from the module is used

cristifalcas:
But since it has the same functionality as pulp-agent and uses the same rpms, I thought that it belongs to this module.

ehelms:
Take a look at -- https://github.com/Craft---/puppet-katello_client  and let me know what you think.

cristifalcas:
I wouldn't used that module in the current form.

We are using jlaska/subscription_manager with those patches:
https://github.com/jlaska/puppet-subscription_manager/pulls

ehelms:
What would that module need updating to use it? I had talked to @Craft--- once about pulling that into our organization at some point when it made sense.

cristifalcas:
I will try to switch our code to use that module and I will do a PR with some changes.

What annoys me most is stuff like "if $katello_client::version == latest" and forcing the katello repo on machines.

ehelms:
Sounds good -- if you get no response from Craft, let me know and we can see about getting it moved here or forking it until we do. I just don't like the idea of tieing 'katello' things to 'pulp' things in this module if we can help it.

ehelms:
@cristifalcas any updates?

cristifalcas:
I didn't have time to do anything regarding this, but considering the previous decision, I consider that this issue should be closed.

PR #97, opened 2015-09-04
Use qpid::tools class

ehelms:
Switches to using qpid::tools class instead of ensuring package
directly to prevent duplicate resource errors

### DIFF
--- a/manifests/broker.pp
+++ b/manifests/broker.pp
@@ -15,7 +15,9 @@ class pulp::broker {
     Service[$broker_service] -> Exec['migrate_pulp_db']
   } else {
     if $pulp::messaging_transport == 'qpid' {
-      package { 'qpid-tools': ensure => installed } -> Class['pulp::service']
+      include ::qpid::tools
+
+      Class['qpid::tools'] -> Class['pulp::service']
     }
   }
 }


manifests/broker.pp
```diff
@@ -15,7 +15,7 @@
     Service[$broker_service] -> Exec['migrate_pulp_db']
   } else {
     if $pulp::messaging_transport == 'qpid' {
-      package { 'qpid-tools': ensure => installed } -> Class['pulp::service']
+      Class['qpid::tools'] -> Class['pulp::service']
```
stbenjam: You gave the example though of qpid being external to the pulp - does it need qpid-tools still on the pulp in that case?

If so, you can `include qpid::tools` here too

ehelms: I think this needs to be present even on a pulp server before migrations are run? I can do further testing to try to pinpoint this better.

stbenjam: Right but `Class['qpid::tools']` isn't including the class, it's just forcing the dependency if it's present.

I think you need to do `include qpid:tools` before this line to account for the case you put qpid on a different server, which doesn't apply to us but could for others.

ehelms:
Good call -- updated

stbenjam:
APT

PR #103, opened 2015-10-02
Remove direct references to Service['qpidd']

ehelms:
The broker.pp handles the relationships for the broker services in
a generic manner. Thus, the requires in the service.pp on the broker
service is unneeded.


stbenjam:
ACK

PR #106, opened 2015-10-07
Drop puppet-foreman in favor of extlib for cache_data

ehelms:


ekohl:
I'm not sure this is a good idea till puppet-foreman has dropped cached_data. Users might still end up with both modules in their environment and the incorrect version could be used.

ehelms:
I think this can be reviewed now that puppet-foreman has been bumped to 5.0.0 in master

ekohl:
:+1:

stbenjam:
ACK

PR #108, opened 2015-10-07
fixes #12094 Added parent param to ensure pulp-nodes-parent is installed

johnpmitsch:


johnpmitsch:
@ehelms I could be missing something, but I can't figure out where to put the test? Is manifest/install.pp a private class?

ehelms:
Introduce a new spec - spec/classes/pulp_install_spec.rb and then look at (https://github.com/Katello/puppet-pulp/blob/master/spec/classes/pulp_config_spec.rb#L18) as an example. This is declaring the pulp main class, however, in your case you'd include the setting of `is_parent` to true.

spec/classes/pulp_install_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+      :operatingsystemmajrelease  => '6.4',
+      :osfamily                   => 'RedHat',
+      :fqdn                       => 'pulp.compony.net',
+      :hostname                   => 'pulp',
+    }
+  end
+
+  describe "with is_parent" do
+    let :facts do
+      :default_facts
```
ehelms: No leading `:`

johnpmitsch:
@ehelms updated, but still not sure why the test isn't passing :/

ehelms:
You'll need to add this as a parameter on `init.pp` for users to be able to set when declaring the pulp class (as well as add a documentation entry)

johnpmitsch:
@ehelms passing :)

manifests/init.pp
```diff
@@ -205,7 +205,8 @@
 # $node_oauth_key::             The oauth key used to authenticate to the parent node
 #
 # $node_oauth_secret::          The oauth secret used to authenticate to the parent node
-#
+# 
+# $is_parent::                  Boolean whether the pulp install is a parent. Defaults to false
```
ehelms: No need to include here what the default is -- params.pp handles that

manifests/init.pp
```diff
@@ -276,6 +277,7 @@
   $node_oauth_effective_user = $pulp::params::node_oauth_effective_user,
   $node_oauth_key            = $pulp::params::node_oauth_key,
   $node_oauth_secret         = $pulp::params::node_oauth_secret,
+  $is_parent                 = $pulp::params::is_parent,
```
ehelms: Can you add a validate_bool for this parameter?

johnpmitsch:
@ehelms updated

stbenjam:
Just my two cents, I would just to call it parent, we don't use the `is_` prefix for any other boolean that I know of.

johnpmitsch:
@stbenjam updated

stbenjam:
Thanks, much appreciated.  Ok from my side.

ehelms:
ACK, thanks @johnpmitsch ! You can go ahead and open PRs to puppet-katello_devel and puppet-katello to set this parameter.

PR #109, opened 2015-10-21
Ensure broker service running before Pulp services and DB manage

ehelms:

PR #110, opened 2015-10-26
bump version to 1.0.3

jlsherrill:


jlsherrill:
@ehelms after chatting with brad, we don't actually need this change for the 2.4 release so we can close it.  The only commit in 1.0.2 is https://github.com/Katello/puppet-pulp/pull/109  which may still be valuable for 2.4.

ehelms:
We don't need to bump with PRs like this -- just release a new module with blacksmith and it will handle.

jlsherrill:
@ehelms yeah @stbenjam filled me here:  https://github.com/Katello/puppet-capsule/pull/63

PR #111, opened 2015-11-19
consolidate node parent params into one parameter

jlsherrill:


ehelms:
LGTM - @stbenjam is this good by you version bumping wise?

stbenjam:
Oh

I would actually suggest you remove both `enable_parent_node`  AND `enable_child_node` and keep parent.  `enable_child_node` looks unused anyway.

I'm pretty opinionated about puppet parameter names, and I am not a fan of prefixes that don't add value like `enable`.  A boolean already implies "enabling" or turning something on and it's clear it's a boolean from the docs and default value.

Also, parent=false implies child so I don't see why it's 2 parameters.

Version-wise I guess you can remove both parameters and bump the major, or deprecate enable_parent_node (see: https://github.com/theforeman/puppet-foreman/blob/2.3-stable/manifests/init.pp#L241) because it's used in one place.

I'm fine if we keep that convention for pulp plugins like "enable_docker", "enable_rpm", etc though.

Thoughts?

ehelms:
I believe a Pulp install can exist without being a Parent or a Child. I think parent implies that it can have child nodes, and child means that it has a parent node. But you could have a stand alone Pulp that isn't intended to be a parent or child (or is later turned into one) AFAIK.

stbenjam:
Really? Why do we have a separate class for children? https://github.com/Katello/puppet-pulp/blob/master/manifests/child.pp

Also, according to pulp docs, there's pulp servers and pulp nodes, so I'm not sure what `enable_parent_node` is supposed to convey? :confused:

jlsherrill:
So there are really two different settings here:
* Do I want my pulp server to be able to act as a pulp child? (i.e. have children sync from it)
* Do I want my pulp server to be able to act as a node child? (i.e. sync from a parent).  This has its own set of settings which is why there is a child class i presume

These are not mutually exclusive, you could have a pulp server be a parent and a child (as eric mentioned).  In the pulp docs they reference 'nodes' and mean children, but you still have to configure the parent properly for it to be able to serve content to its children.

stbenjam:
Ok, so this is ok for me, to remove parent and bump the major.  Can we fix the enable_child_node param in here too so it does something?

jlsherrill:
@stbenjam or just remove it altogether?

stbenjam:
That works too

jlsherrill:
@stbenjam updated to remove enable_child_node

stbenjam:
ACK from me

ehelms:
ACK for me

ekohl:
Thanks!

PR #112, opened 2016-01-22
fixes #13189 - allows authentication with username and certificate

cfouant:


ehelms:
MInd changing the commit message a bit to not include capsules since they are not part of Pulp itself? That way it better captures what this enables with respect to Pulp access.

cfouant:
@ehelms - updated commit message, thanks

templates/etc/httpd/conf.d/_ssl_username.conf.erb
```diff
@@ -0,0 +1,5 @@
+#
+# Apache configuration file for Pulp's Puppet support
```
ehelms: This seems incorrect documentation, I think you'd  be OK reducing this to a one line comment above the directive.

ehelms:
ACK - thanks @cfouant !

PR #113, opened 2016-01-26
Fixes #13394 - Added xsendfile support for pulp 2.8

parthaa:


manifests/xsendfile.pp
```diff
@@ -0,0 +1,3 @@
+class pulp::xsendfile {
+  include ::apache::mod::xsendfile
```
ehelms: This should just go in the apache.pp -- https://github.com/Katello/puppet-pulp/blob/master/manifests/apache.pp#L5

parthaa:
@ehelms updated

ehelms:
ACK - thanks @parthaa !

PR #114, opened 2016-01-26
fix template :undef checks and yum proxy variables

walkamongus:
I was getting inconsistent template results using the explicit checks for :undef -- some values were correctly populated and other weren't. I was also unable to set yum proxy settings because the variables were out of scope.

This PR fixes both issues.


ekohl:
You're right there should be better checking, but I'm a bit unsure if this also works as expected with empty strings and across various versions. I think the best way is to add some tests.

walkamongus:
Should any of these ever be an empty string? If not, then I think there should be a check for that in the parameter validations in init.pp, yes? Validation there could catch any empty strings as well as non-string values.

PR #115, opened 2016-01-27
Mongodb fact refactor

walkamongus:
Here is a proposed fix for the following issues:
* executing repoquery on a box without yum-utils causes `sh: repoquery: command not found` to be output
* if STDERR is not redirected for the `yum info` command,  the command outputs `Error: No matching Packages to list`

### DIFF
--- a/lib/facter/facts.rb
+++ b/lib/facter/facts.rb
@@ -2,7 +2,7 @@ Facter.add(:mongodb_version) do
   setcode do
     commands = ["rpmquery --qf='%{version}-%{release}' mongodb",
                 "repoquery --qf='%{version}-%{release}' mongodb",
-                %[LC_ALL=en_US yum -e 0 -d 0 info mongodb | awk '/^Version/ { version=$3; } /^Release/ { print version "-" $3; exit }']]
+                %[LC_ALL=en_US yum -e 0 -d 0 info mongodb 2>&1 | awk '/^Version/ { version=$3; } /^Release/ { print version "-" $3; exit }']]
     ret = nil
     commands.each do |command|
       version = `#{command}`.chomp
--- a/lib/facter/facts.rb
+++ b/lib/facter/facts.rb
@@ -10,7 +10,7 @@ Facter.add(:mongodb_version) do
     commands.each do |command|
       exe = command.partition(' ').first
       if system("which #{exe} > /dev/null 2>&1")
-        version = `#{command}`.chomp
+        version = `#{command} 2> /dev/null`.chomp
         if $?.success? && !version.empty?
           ret = version
           break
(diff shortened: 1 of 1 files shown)


ehelms:
Looks OK to me, @iNecas if you have a moment to inspect it, would be appreciated.

ehelms:
@iNecas trying you again just in case :) if not, I'll try and work on this

iNecas:
@ehelms sry, no resource to look at this right now :(

stbenjam:
When I try this on Fedora, the yum deprecation warning isn't surpressed:

```
 stbenjam  ~  bar  FACTERLIB=$PWD facter mongodb_version

Yum-utils package has been deprecated, use dnf instead.
See 'man yum2dnf' for more information.


2.6.11-1.fc22
```

I'm not sure what's going on there, the command actually run does suppress the output. :confused:

stbenjam:
Ah, I see, it's coming from repoquery not yum, as it's deprecated now in favor of using dnf.

Either redirect it's stderr:
```shell
repoquery --qf='%{version}-%{release}' mongodb 2>/dev/null
```

Or maybe just redirect it all in the line below:

```
version = `#{command} 2>/dev/null`.chomp
```

TJM:
FYI, `which` is not guaranteed to be installed either. I found *that* out messing with CentOS 7 docker images. Minimal installs are actually starting to get "minimal" :)

Is there some reason we aren't just going with `rpm -q --qf='%{version}-%{release}' mongodb` ? running repoquery or rpmquery would be like calling some subcommand of git directly. Also, calling "yum info" may have the nasty habit of showing you information for a package that is not yet installed...

```
[tommy@lark-sensu-01 ~]$ rpm -q mongodb
package mongodb is not installed
[tommy@lark-sensu-01 ~]$ yum -e 0 -d 0 info mongodb 2>&1 | awk '/^Version/ { version=$3; } /^Release/ { print version "-" $3; exit }'
2.6.11-1.el7
[tommy@lark-sensu-01 ~]$
```

Which lists a version that is *available* to be installed, but I do not think that is the intended operation.

~tommy

stbenjam:
```
Which lists a version that is available to be installed, but I do not think that is the intended operation.
```

It's intended.  The first option rpmquery is basically the same as rpm -q, which shows the installed package version, but in case it's not installed, we need to know what version will be installed.  It's a hack basically, because puppet doesn't let us look at the version of a package it installs until the second run, but we need to know on the first run so we can complete the configuration in the first pass.

TJM:
For those of us configured to run puppet via cron, this is particularly annoying. What needs to happen to merge this?

Tommy

stbenjam:
From my perspective it's good to go once my last comment is addressed.

walkamongus:
I just pushed up redirection of command errors to /dev/null per comments. Let me know if you need anything squashed or rebased!

ekohl:
@stbenjam mind having a look again?

stbenjam:
Ah sorry for the delay, looks good to me and tests well - merging!

Thanks @walkamongus!

PR #116, opened 2016-01-28
Fixes #13431 - Apache changes for pulp 2.8

parthaa:
Includes updates for docker/nodes/ostree/puppet and rpm

### DIFF
--- a/manifests/apache.pp
+++ b/manifests/apache.pp
@@ -72,6 +72,8 @@ class pulp::apache {
   }
 
   if $pulp::manage_httpd or $pulp::manage_plugins_httpd {
+    pulp::apache_plugin {'content' : vhosts80 => false}
+
     file { '/etc/pulp/vhosts80/':
       ensure => directory,
       owner  => 'apache',
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -31,6 +31,9 @@
 #
 # $broker_use_ssl::             Require SSL if set to 'true', otherwise do not require SSL.
 #
+# $tasks_login_method::         Select the SASL login method used to connect to the broker. This should be left
+#                               unset except in special cases such as SSL client certificate authentication.
+#
 # $ca_cert::                    full path to the CA certificate that will be used to sign consumer
 #                               and admin identification certificates; this must match the value of
 #                               SSLCACertificateFile in /etc/httpd/conf.d/pulp.conf
@@ -64,6 +67,14 @@
 #                               certificates, which are used to validate certificates passed from the other end
 #                               of the connection.
 #
+# $db_unsafe_autoretry:         If true, retry commands to the database if there is a connection error.
+#                               Warning: if set to true, this setting can result in duplicate records.
+#
+# $db_write_concern:            Write concern of 'majority' or 'all'. When 'all' is specified, 'w' is set to
+#                               number of seeds specified. For version of MongoDB < 2.6, replica_set must also
+#                               be specified. Please note that 'all' will cause Pulp to halt if any of the
+#                               replica set members is not available. 'majority' is used by defau
+#
 # $server_name::                hostname the admin client and consumers should use when accessing
 #                               the server; if not specified, this is defaulted to the server's hostname
 #
@@ -75,6 +86,8 @@
 #
 # $log_level::                  The desired logging level. Options are: CRITICAL, ERROR, WARNING, INFO, DEBUG,
 #                               and NOTSET. Pulp will default to INFO.
+#
+# $server_working_directory::   Path to where pulp workers can create working directories needed to complete tasks
 #
 # $rsa_key::                    The RSA private key used for authentication.
 #
@@ -133,6 +146,10 @@
 # $messaging_topic_exchange::   The name of the exchange to use. The exchange must be a topic exchange. The
 #                               default is 'amq.topic', which is a default exchange that is guaranteed to exist on a Qpid broker.
 #
+# $messaging_event_notifications_enabled:: Enables or disables Pulp event notfications on the message bus. Defaults to 'false'.
+#
+# $messaging_event_notification_url:: The AMQP URL for event notifications. Defaults to 'qpid://localhost:5672/'.
+#
 # $email_host::                 host name of the MTA pulp should relay through
 #
 # $email_port::                 destination port to connect on
@@ -141,6 +158,29 @@
 #
 # $email_enabled::              boolean controls whether or not emails will be sent
 #                               type:boolean
+#
+# $lazy_enabled::               boolean controls whether redirect is enabled
+#
+# $lazy_redirect_host::         The host FQDN or IP to which requests are redirected.
+#
+# $lazy_redirect_port::         The TCP port to which requests are redirected
+#
+# $lazy_redirect_path::         The base path to which requests are redirected
+#
+# $lazy_https_retrieval::       boolean; controls whether Pulp uses HTTPS or HTTP to
+#                               retrieve content from the streamer.
+#                               WARNING: Setting this to 'false' is not safe if you wish
+#                               to use Pulp to provide repository entitlement
+#                               enforcement. It is strongly recommended to keep
+#                               this set to 'true' and use certificates that are
+#                               signed by a trusted authority on the web server
+#                               that serves as the streamer reverse proxy.
+#
+# $lazy_download_interval::     The interval in minutes between checks for content cached
+#                               by the Squid proxy.
+#
+# $lazy_download_concurrency:   The number of downloads to perform concurrently when
+#                               downloading content from the Squid cache.
 #
 # $proxy_url::                  URL of the proxy server
 #
@@ -223,6 +263,8 @@ class pulp (
   $db_ssl_certfile           = $pulp::params::db_ssl_certfile,
   $db_verify_ssl             = $pulp::params::db_verify_ssl,
   $db_ca_path                = $pulp::params::db_ca_path,
+  $db_unsafe_autoretry       = $pulp::params::db_unsafe_autoretry,
+  $db_write_concern          = $pulp::params::db_write_concern,
   $server_name               = $pulp::params::server_name,
   $key_url                   = $pulp::params::key_url,
   $ks_url                    = $pulp::params::ks_url,
@@ -230,6 +272,7 @@ class pulp (
   $default_password          = $pulp::params::default_password,
   $debugging_mode            = $pulp::params::debugging_mode,
   $log_level                 = $pulp::params::log_level,
+  $server_working_directory  = $pulp::params::server_working_directory,
   $rsa_key                   = $pulp::params::rsa_key,
   $rsa_pub                   = $pulp::params::rsa_pub,
   $ca_cert                   = $pulp::params::ca_cert,
@@ -250,12 +293,22 @@ class pulp (
   $messaging_ca_cert         = $pulp::params::messaging_ca_cert,
   $messaging_client_cert     = $pulp::params::messaging_client_cert,
   $messaging_topic_exchange  = $pulp::params::messaging_topic_exchange,
+  $messaging_event_notifications_enabled = $pulp::params::messaging_event_notifications_enabled,
+  $messaging_event_notification_url = $pulp::params::messaging_event_notification_url,
   $broker_url                = $pulp::params::broker_url,
   $broker_use_ssl            = $pulp::params::broker_use_ssl,
+  $tasks_login_method        = $pulp::params::tasks_login_method,
   $email_host                = $pulp::params::email_host,
   $email_port                = $pulp::params::email_port,
   $email_from                = $pulp::params::email_from,
   $email_enabled             = $pulp::params::email_enabled,
+  $lazy_enabled              = $pulp::params::lazy_enabled,
+  $lazy_redirect_host        = $pulp::params::lazy_redirect_host,
+  $lazy_redirect_port        = $pulp::params::lazy_redirect_port,
+  $lazy_redirect_path        = $pulp::params::lazy_redirect_path,
+  $lazy_https_retrieval      = $pulp::params::lazy_https_retrieval,
+  $lazy_download_interval    = $pulp::params::lazy_download_interval,
+  $lazy_download_concurrency = $pulp::params::lazy_download_concurrency,
   $consumers_crl             = $pulp::params::consumers_crl,
   $reset_cache               = $pulp::params::reset_cache,
   $ssl_verify_client         = $pulp::params::ssl_verify_client,
@@ -298,6 +351,10 @@ class pulp (
   validate_bool($enable_parent_node)
   validate_bool($repo_auth)
   validate_bool($reset_cache)
+  validate_bool($db_unsafe_autoretry)
+  validate_bool($messaging_event_notifications_enabled)
+  validate_bool($lazy_enabled)
+  validate_bool($lazy_https_retrieval)
   validate_array($disabled_authenticators)
   validate_hash($additional_wsgi_scripts)
 
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -13,12 +13,15 @@ class pulp::params {
   $db_ssl_certfile = undef
   $db_verify_ssl = true
   $db_ca_path = '/etc/pki/tls/certs/ca-bundle.crt'
+  $db_unsafe_autoretry = false
+  $db_write_concern = undef
 
   $server_name = downcase($::fqdn)
   $key_url = '/pulp/gpg'
   $ks_url = '/pulp/ks'
   $debugging_mode = false
   $log_level = 'INFO'
+  $server_working_directory = undef
 
   $rsa_key = '/etc/pki/pulp/rsa.key'
   $rsa_pub = '/etc/pki/pulp/rsa_pub.key'
@@ -38,9 +41,12 @@ class pulp::params {
   $messaging_ca_cert = undef
   $messaging_client_cert = undef
   $messaging_topic_exchange = 'amq.topic'
+  $messaging_event_notifications_enabled = false
+  $messaging_event_notifications_url = undef
 
   $broker_url = "qpid:///guest@${::fqdn}:5672"
   $broker_use_ssl = false
+  $tasks_login_method = undef
 
   $ca_cert = '/etc/pki/pulp/ca.crt'
   $ca_key = '/etc/pki/pulp/ca.key'
@@ -61,6 +67,14 @@ class pulp::params {
   $email_port = 25
   $email_from = "no-reply@${::domain}"
   $email_enabled = false
+
+  $lazy_enabled = false
+  $lazy_redirect_host = downcase($::fqdn)
+  $lazy_redirect_port = undef
+  $lazy_redirect_path = '/streamer/'
+  $lazy_https_retrieval = false
+  $lazy_download_interval = 30
+  $lazy_download_concurrency = 5
 
   $consumers_crl = undef
 
--- a/spec/classes/pulp_apache_spec.rb
+++ b/spec/classes/pulp_apache_spec.rb
@@ -69,7 +69,7 @@ describe 'pulp::apache' do
       let :pre_condition do
         "class {'pulp': manage_httpd => false, manage_plugins_httpd => true, enable_rpm => true}"
       end
- 
+
        it { should_not contain_apache__vhost('pulp-http') }
        it { is_expected.to contain_file('/etc/httpd/conf.d/pulp_rpm.conf') }
     end
@@ -91,6 +91,39 @@ describe 'pulp::apache' do
       end
     end
 
+    describe 'with manage_httpd true or manage_plugins_httpd true' do
+      let :pre_condition do
+        "class {'pulp': manage_httpd => true}"
+      end
+
+      it 'should configure pulp_content' do
+        is_expected.to contain_file('/etc/httpd/conf.d/pulp_content.conf').with(
+        :content => 'WSGISocketPrefix run/wsgi
+WSGIProcessGroup pulp-content
+WSGIApplicationGroup pulp-content
+WSGIScriptAlias /pulp/content /usr/share/pulp/wsgi/content.wsgi
+WSGIDaemonProcess pulp-content user=apache group=apache processes=3 display-name=%{GROUP}
+WSGIImportScript /usr/share/pulp/wsgi/content.wsgi process-group=pulp-content application-group=pulp-content
+
+<Files content.wsgi>
+    WSGIPassAuthorization On
+    WSGIProcessGroup pulp-content
+    WSGIApplicationGroup pulp-content
+    SSLRenegBufferSize  1048576
+    SSLVerifyDepth 9
+    SSLOptions +StdEnvVars +ExportCertData
+    SSLVerifyClient require
+</Files>
+
+<Location /pulp/content/>
+    XSendFile on
+    XSendFilePath /var/lib/pulp/content
+    XSendFilePath /var/lib/pulp/published
+</Location>
+')
+      end
+    end
+
     describe 'with enable_rpm' do
       let :pre_condition do
         "class {'pulp': enable_rpm => true}"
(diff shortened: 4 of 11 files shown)


templates/etc/httpd/conf.d/pulp_ostree.conf.erb
```diff
@@ -7,10 +7,10 @@
...
 <Directory /var/www/pub/ostree>
-    WSGIAccessScript /srv/pulp/repo_auth.wsgi
+    WSGIAccessScript /usr/share/pulp/wsgi/repo_auth.wsgi
     SSLRequireSSL
     SSLVerifyClient <%= scope['pulp::ssl_verify_client'] %>
     SSLVerifyDepth 2
     SSLOptions +StdEnvVars +ExportCertData +FakeBasicAuth
     Options FollowSymLinks Indexes
-</Directory>
+</Directory>
```
ekohl: Nitpick: Keep the newline?

templates/etc/httpd/conf.d/pulp_puppet.conf.erb
```diff
@@ -36,8 +36,10 @@ Alias /pulp/puppet_files /var/www/pub/puppet/files
 # and so the following redirect will match any path that isn't the above.
 RedirectMatch ^\/?pulp_puppet\/forge\/[^\/]+\/[^\/]+\/(?!api\/v1\/releases\.json)(.*)$ /$1
 
-# for puppet < 3.3
-WSGIScriptAlias /api/v1 /srv/pulp/puppet_forge_pre33_api.wsgi
-# for puppet >= 3.3
-WSGIScriptAlias /pulp_puppet/forge /srv/pulp/puppet_forge_post33_api.wsgi
+WSGIDaemonProcess pulp_forge user=apache group=apache processes=3 display-name=%{GROUP}
```
ekohl: Should the number of processes be configurable?

ekohl:
I don't know pulp very well so you may be better able to answer this: how backwards compatible is this with older pulp versions?

beav:
@ekohl These changes require Pulp 2.8 or newer. For example,

```
-    WSGIAccessScript /srv/pulp/repo_auth.wsgi
+    WSGIAccessScript /usr/share/pulp/wsgi/repo_auth.wsgi
```
The path to that wsgi script changed in Pulp 2.8, as part of https://github.com/pulp/pulp/commit/73b813b6c2ce682708f9386d71a29393e6d8fe77.

ekohl:
In that case this is a major version bump for the module. That's fine and I don't think we need parameters to support both but I'll look at @ehelms for a definitive answer :)

beav:
@parthaa the added test should work in theory, I am looking into it now

spec/classes/pulp_apache_spec.rb
```diff
@@ -259,6 +290,35 @@
...
+
+<Directory /var/www/pub/ostree>
+    WSGIAccessScript /usr/share/pulp/wsgi/repo_auth.wsgi
+    SSLRequireSSL
+    SSLVerifyClient require
+    SSLVerifyDepth 2
+    SSLOptions +StdEnvVars +ExportCertData +FakeBasicAuth
+    Options FollowSymLinks Indexes
+</Directory>
+
```
beav: @parthaa this newline in the file comparison is causing the breakage :hamster:

beav:
@parthaa 

i also added this test to pulp_admin_spec.rb ensure the admin extension gets installed:

```ruby
    context 'install with ostree param' do
      let(:params) do {
          'enable_ostree' => true,
        } end

      it { should contain_package('pulp-ostree-admin-extensions').with_ensure('installed') }
    end
```

parthaa:
@beav thanks for the pointers

manifests/apache.pp
```diff
@@ -71,6 +71,9 @@
   }
 
   if $pulp::manage_httpd or $pulp::manage_plugins_httpd {
+
+    pulp::apache_plugin {'content' : vhosts80 => false}
```
ehelms: No space before the `:`

templates/etc/httpd/conf.d/pulp_docker.conf.erb
```diff
@@ -4,8 +4,21 @@
...
-<Location /var/www/pub/docker/web>
+# Docker v2
+Alias /pulp/docker/v2 /var/www/pub/docker/v2/web
+<Directory /var/www/pub/docker/v2/web>
+    Header set Docker-Distribution-API-Version "registry/2.0"
+    SSLRequireSSL
+    Options FollowSymlinks Indexes
+</Directory>
+
+# Docker v1
```
ehelms: 2.8 does support Docker v1?

templates/etc/httpd/conf.d/pulp_nodes.conf.erb
```diff
@@ -22,3 +22,14 @@ Alias /pulp/nodes/https/repos /var/www/pulp/nodes/https/repos
   SSLOptions +FakeBasicAuth
   SSLRequire %{SSL_CLIENT_S_DN_O} eq "PULP" and %{SSL_CLIENT_S_DN_OU} eq "NODES"
 </Directory>
+
+Alias /pulp/nodes/content /var/www/pulp/nodes/content
+
+<Directory /var/www/pulp/nodes/content >
+  Options FollowSymLinks Indexes
+  SSLRequireSSL
+  SSLVerifyClient require
```
ehelms: Looks like this should follow https://github.com/Katello/puppet-pulp/pull/116/files#diff-56c104ac5f76e0676fddcfb2da13dd62R20

templates/etc/httpd/conf.d/pulp_rpm.conf.erb
```diff
@@ -22,12 +37,13 @@ Alias /pulp/exports /var/www/pub/yum/https/exports
     Options FollowSymLinks Indexes
 </Directory>
 
-
 # -- HTTPS ISOS
 Alias /pulp/isos /var/www/pub/https/isos
 
 <Directory /var/www/pub/https/isos>
-    WSGIAccessScript /srv/pulp/repo_auth.wsgi
+
```
ehelms: This blank line can be removed

templates/server.conf.erb
```diff
@@ -55,6 +65,8 @@ ssl_certfile: <%= scope['pulp::db_ssl_certfile'] %>
 verify_ssl: <%= scope['pulp::db_verify_ssl'] %>
 ca_path: <%= scope['pulp::db_ca_path'] %>
 <% end -%>
+# unsafe_autoretry: false
+# write_concern: majority
```
ehelms: These should be added as parameters, otherwise there will be no way to configure them (because puppet would overwrite any manual changes to this file).

templates/server.conf.erb
```diff
@@ -78,7 +92,7 @@ default_login: <%= scope['pulp::default_login'] %>
 default_password: <%= scope['pulp::default_password'] %>
 debugging_mode: <%= scope['pulp::debugging_mode'] %>
 log_level: <%= scope['pulp::log_level'] %>
-
+# working_directory: /var/cache/pulp
```
ehelms: Need a parameter

templates/server.conf.erb
```diff
@@ -271,7 +295,8 @@ cacert: <%= scope['pulp::messaging_ca_cert'] %>
 clientcert: <%= scope['pulp::messaging_client_cert'] %>
 <% end -%>
 topic_exchange: '<%= scope['pulp::messaging_topic_exchange'] %>'
-
+# event_notifications_enabled: false
+# event_notification_url: qpid://localhost:5672/
```
ehelms: Need parameters

templates/server.conf.erb
```diff
@@ -313,7 +340,7 @@ cacert: <%= scope['pulp::messaging_ca_cert'] %>
 keyfile: <%= scope['pulp::messaging_client_cert'] %>
 certfile: <%= scope['pulp::messaging_client_cert'] %>
 <% end -%>
-
+# login_method:
```
ehelms: Needs parameter

templates/server.conf.erb
```diff
@@ -338,3 +365,46 @@ host: <%= scope['pulp::email_host'] %>
...
+#   downloading content from the Squid cache.
+
+[lazy]
+# enabled: false
+# redirect_host:
+# redirect_port:
+# redirect_path:
+# https_retrieval: true
+# download_interval: 30
+# download_concurrency: 5
```
ehelms: These all need parameters to configure them

templates/etc/httpd/conf.d/pulp_docker.conf.erb
beav: My understanding is that in Pulp's realm, repos are not associated with any particular content type. Thus, you can mix v1 and v2 content in the same repo even though they work with different units (images vs blobs).

In practice, it would be extremely complicated to support this type of arrangement, but IMO the puppet module may as well expose the directory for users of this puppet module besides Katello.

templates/server.conf.erb
```diff
@@ -338,3 +365,46 @@ host: <%= scope['pulp::email_host'] %>
...
+# download_interval:
+#   The interval in minutes between checks for content cached
+#   by the Squid proxy.
+#
+# download_concurrency:
+#   The number of downloads to perform concurrently when
+#   downloading content from the Squid cache.
+
+[lazy]
+# enabled: false
```
beav: If this is set to 'enabled: false' currently, would it harm anything to uncomment this section? I think it would be better to either uncomment, or remove for now and put it into a different PR if we plan on enabling it later.

templates/server.conf.erb
jlsherrill: so i had discussed this a bit with partha.  there are existing settings that do not have parameters and are commented out (https://github.com/Katello/puppet-pulp/blob/master/templates/server.conf.erb#L166-L206), and partha wanted to make this file look as much like the stock server.conf as possible hence these entries.  I assumed we could add parameters for all of them either in a followup PR or as needed.   Ideally every config value would be configurable but that already isn't the case, so is it that we prefer:

a) any new config entires added should have parameters and should be 100% up to date with the corresponding pulp stock configs
b) config entires shouldn't be added until there are parameters, meaning some may be missing
c) $$$$

Trying to understand myself for future contributions :)

templates/server.conf.erb
parthaa: Was planning to do them with the lazy sync work.. More here for completeness

templates/server.conf.erb
ehelms: Agreed, the data_repeaing was overlooked, and the ldap, to be fair, is deprecated. So, I would vote for adding them when the parameters are added. As that avoids missing them in the future, and its fairly easy to do.

jlsherrill: I'm okay with this, thanks

manifests/apache.pp
```diff
@@ -3,7 +3,6 @@
   include ::apache
   include ::apache::mod::wsgi
   include ::apache::mod::ssl
-  include ::apache::mod::xsendfile
```
jlsherrill: don't think this was intended

manifests/params.pp
```diff
@@ -62,6 +68,14 @@
   $email_from = "no-reply@${::domain}"
   $email_enabled = false
 
+  $lazy_enabled = undef
+  $lazy_redirect_host = undef
+  $lazy_redirect_port = undef
+  $lazy_redirect_path = undef
+  $lazy_https_retrieval = undef
+  $lazy_download_interval = undef
+  $lazy_download_concurrency = undef
```
jlsherrill: I think all of these should have default value, whever the pulp server.conf default is

parthaa: Well they are commented in that file

templates/server.conf.erb
```diff
@@ -338,3 +379,60 @@ host: <%= scope['pulp::email_host'] %>
...
+<% end -%>
+<% if scope['pulp::lazy_https_retrieval'] != :undef -%>
+https_retrieval: <%= scope['pulp::lazy_https_retrieval'] %>
+<% end -%>
+<% if scope['pulp::lazy_download_interval'] != :undef -%>
+download_interval: <%= scope['pulp::lazy_download_interval'] %>
+<% end -%>
+<% if scope['pulp::lazy_download_concurrency'] != :undef -%>
+download_concurrency: <%= scope['pulp::lazy_download_concurrency'] %>
+<% end -%>
```
jlsherrill: i think it would make more sense to leave out the others if lazy_enabled is false and include them all (with sane defaults) if its is enabled.

manifests/params.pp
parthaa: The way I saw the pattern
```
<% if scope['pulp::lazy_enabled'] != :undef -%>
enabled: <%= scope['pulp::lazy_enabled'] %>
<% end -%>

```

jlsherrill: I think that is only appropriate if the 'default' is '' or nil
if there is a default value (i.e. false) lets juts use that

jlsherrill: unless i'm off base here, @ehelms please correct me if i am

templates/server.conf.erb
```diff
@@ -338,3 +379,48 @@ host: <%= scope['pulp::email_host'] %>
...
+
+[lazy]
+enabled: <%= scope['pulp::lazy_enabled'] %>
+<% if scope['pulp::lazy_enabled'] -%>
+redirect_host: <%= scope['pulp::lazy_redirect_host'] %>
+redirect_port: <%= scope['pulp::lazy_redirect_port'] %>
+redirect_path: <%= scope['pulp::lazy_redirect_path'] %>
+https_retrieval: <%= scope['pulp::lazy_https_retrieval'] %>
+download_interval: <%= scope['pulp::lazy_download_interval'] %>
+download_concurrency: <%= scope['pulp::lazy_download_concurrency'] %>
```
ehelms: Will Pulp function properly if any of these are set to nil/null ? Or are some of these required if this is enabled?

templates/server.conf.erb
```diff
@@ -313,7 +349,12 @@ cacert: <%= scope['pulp::messaging_ca_cert'] %>
 keyfile: <%= scope['pulp::messaging_client_cert'] %>
 certfile: <%= scope['pulp::messaging_client_cert'] %>
 <% end -%>
-
+<% if scope['pulp::messaging_ca_cert'] -%>
+cacert: <%= scope['pulp::messaging_ca_cert'] %>
+<% end -%>
```
ehelms: This exists on line 345

templates/server.conf.erb
```diff
@@ -313,7 +349,12 @@ cacert: <%= scope['pulp::messaging_ca_cert'] %>
...
+<% if scope['pulp::tasks_login_method'] != :undef -%>
```
ehelms: The `!= :undef` is not required and won't actually do anything.

ehelms: True of other spots in the file as well.

manifests/init.pp
```diff
@@ -223,6 +263,8 @@
   $db_ssl_certfile           = $pulp::params::db_ssl_certfile,
   $db_verify_ssl             = $pulp::params::db_verify_ssl,
   $db_ca_path                = $pulp::params::db_ca_path,
+  $db_unsafe_autoretry       = $pulp::params::db_unsafe_autoretry,
```
ehelms: Can you add a validate_bool for this parameter?

manifests/init.pp
```diff
@@ -250,12 +293,22 @@
   $messaging_ca_cert         = $pulp::params::messaging_ca_cert,
   $messaging_client_cert     = $pulp::params::messaging_client_cert,
   $messaging_topic_exchange  = $pulp::params::messaging_topic_exchange,
+  $messaging_event_notifications_enabled = $pulp::params::messaging_event_notifications_enabled,
```
ehelms: Can you add a validate_bool for this parameter?

manifests/init.pp
```diff
@@ -250,12 +293,22 @@
...
+  $messaging_event_notification_url = $pulp::params::messaging_event_notification_url,
   $broker_url                = $pulp::params::broker_url,
   $broker_use_ssl            = $pulp::params::broker_use_ssl,
+  $tasks_login_method        = $pulp::params::tasks_login_method,
   $email_host                = $pulp::params::email_host,
   $email_port                = $pulp::params::email_port,
   $email_from                = $pulp::params::email_from,
   $email_enabled             = $pulp::params::email_enabled,
+  $lazy_enabled              = $pulp::params::lazy_enabled,
```
ehelms: Can you add a validate_bool for this parameter?

manifests/params.pp
```diff
@@ -38,9 +41,12 @@
   $messaging_ca_cert = undef
   $messaging_client_cert = undef
   $messaging_topic_exchange = 'amq.topic'
+  $messaging_event_notifications_enabled = undef
```
ehelms: This should probably be false since its a boolean.

manifests/params.pp
```diff
@@ -13,12 +13,15 @@
   $db_ssl_certfile = undef
   $db_verify_ssl = true
   $db_ca_path = '/etc/pki/tls/certs/ca-bundle.crt'
+  $db_unsafe_autoretry = undef
```
ehelms: This should probably be false since its a boolean.

manifests/init.pp
```diff
@@ -250,12 +293,22 @@
...
+  $lazy_redirect_host        = $pulp::params::lazy_redirect_host,
+  $lazy_redirect_port        = $pulp::params::lazy_redirect_port,
+  $lazy_redirect_path        = $pulp::params::lazy_redirect_path,
+  $lazy_https_retrieval      = $pulp::params::lazy_https_retrieval,
```
ehelms: Same here -- add a validate_bool

templates/server.conf.erb
beav: @ehelms I think it will rely on these defaults if anything is nil/null:

```python
    'lazy': {
        'enabled': 'false',
        'redirect_host': socket.getfqdn(),
        'redirect_port': '',
        'redirect_path': '/streamer/',
        'https_retrieval': 'true',
        'download_interval': '30',
        'download_concurrency': '5'
    },
```

I assume these are sane for Pulp to function :hamster: If not, it's a Pulp bug IMO.

beav:
This patch looks OK to me, aside from the fixups suggested by @ehelms.

templates/server.conf.erb
jeremycline: These should be sane defaults for some definition of sane. They assume the Apache reverse proxy is running on localhost (`'redirect_host': socket.getfqdn()`) and that it has a certificate that is signed by a CA that the Pulp server host trusts. You could set `https_retrieval: false` if the pulp server, the reverse proxy, squid, and the pulp streamer are all on the same host.

beav: @jeremycline those assumptions sound reasonable to me and should work ok I think, thanks! :smile_cat:

jeremycline: @beav :+1:

parthaa:
@ehelms @beav  @jlsherrill  updated :). Thanks!

jlsherrill:
Getting an error:

Jan 29 19:06:34 katello-2-8-test pulp: pulp.server.webservices.application:ERROR: (16764-09536) PulpCodedException: Database 'write_concern' config can only be 'majority' or 'all'. Refer to /etc/pulp/server.conf for proper use.

jlsherrill:
it actually looks like those undefs were needed:
```
working_directory: undef
```

some examples:

```
[database]
name: pulp_database
seeds: localhost:27017
username: undef
password: undef
replica_set: undef
ssl: false
write_concern: undef
```

parthaa:
Ugh :/

templates/server.conf.erb
```diff
@@ -33,36 +37,48 @@
...
+#                    number of seeds specified. For version of MongoDB < 2.6, replica_set must also
+#                    be specified. Please note that 'all' will cause Pulp to halt if any of the
+#                    replica set members is not available. 'majority' is used by defau
 
 [database]
-name: <%= scope['pulp::db_name'] %>
+name: <%= scope['pulp::db_name']%>
 seeds: <%= scope['pulp::db_seeds'] %>
-<% if scope['pulp::db_username'] != :undef -%>
+<% if scope['pulp::db_username'] != undef -%>
```
jlsherrill: i think you left out the ':'  a bunch of these didn't need updates, find/replace error?

parthaa: Thx, updated

templates/server.conf.erb
```diff
@@ -46,7 +56,7 @@ password: <%= scope['pulp::db_password'] %>
 <% if scope['pulp::db_replica_set'] != :undef -%>
 replica_set: <%= scope['pulp::db_replica_set'] %>
 <% end -%>
-ssl: <%= scope['pulp::db_ssl'] %>
+ssl: <%= scope['pulp::db_ssl']%>
```
jlsherrill: removed the space :/ on this line and ~10 lines above

templates/server.conf.erb
```diff
@@ -55,14 +65,20 @@ ssl_certfile: <%= scope['pulp::db_ssl_certfile'] %>
 verify_ssl: <%= scope['pulp::db_verify_ssl'] %>
 ca_path: <%= scope['pulp::db_ca_path'] %>
 <% end -%>
-
+<% if scope['pulp::db_unsafe_autoretry'] -%>
+unsafe_autoretry: <%= scope['pulp::db_unsafe_autoretry'] %>
+<% end -%>
+<% if scope['pulp::db_write_concern'] != :undef-%>
```
ehelms: As I understand it, given undef is not a real puppet concept, the best way to do these checks is:

https://github.com/theforeman/puppet-foreman_proxy/blob/master/templates/dhcp.yml.erb#L22

templates/etc/httpd/conf.d/pulp_rpm.conf.erb
```diff
@@ -2,17 +2,32 @@
 # Apache configuration file for pulp web services and repositories
 #
 
-# -- HTTPS Repositories ---------
-Alias /pulp/repos /var/www/pub/yum/https/repos
+AddType application/x-pkcs7-crl .crl
+AddType application/x-x509-ca-cert .crt
```
ehelms: Is this needed here? I saw it also in https://github.com/Katello/puppet-katello/pull/98/files#diff-40567eb5f7aff653485984fef811556cR9 or maybe the question is the other way around?

ehelms:
All the path changes, in the httpd conf files, those are backwards incompatible I assume?

jlsherrill:
@ehelms yes as @ekohl brought up here:  https://github.com/Katello/puppet-pulp/pull/116#issuecomment-176197701  this will need a major version bump

templates/etc/httpd/conf.d/pulp_rpm.conf.erb
jlsherrill: I'm assuming these were in the pulp stock configs @parthaa ?

templates/server.conf.erb
parthaa: Yet another undef nightmare :/

templates/etc/httpd/conf.d/pulp_rpm.conf.erb
parthaa: @jlsherrill @ehelms yes they were in pulp stock configs

templates/server.conf.erb
parthaa: @ehelms Cant I argue that in our params.pp  we are exclusively marking it "undef" , why do i need to check for each
```
 unless [nil, :undefined, :undef, ''].
```

templates/etc/httpd/conf.d/pulp_rpm.conf.erb
beav: I don't know of anyone who's downloading CAs or CRLs from this location but a Pulp person would know for sure.

It doesn't hurt anything to have this in here so I say just leave it :hamster:

parthaa:
@ehelms @jlsherrill updated the undef checks

ehelms:
ACK

ehelms:
Thanks @parthaa !

PR #117, opened 2016-01-28
fix template :undef checks and yum proxy variables

walkamongus:
Supercedes #114


ehelms:
Generally looks OK to me. Could you squash into a single commit?

walkamongus:
I will squash. There was discussion in #114 regarding whether this will handle empty strings or not. I believe Puppet < 4.0.0 equates an empty string to undef but in > 4.0.0, empty strings are true. If parameters should be strings and not empty, this should be part of the parameter validations in init.pp.

I can add those validations in if it is desired.

ehelms:
You'll need a rebase here after it was updated for Pulp 2.8

walkamongus:
rebased and converted remaining undef check to the "new" method

ehelms:
ACK - thanks @walkamongus !

PR #122, opened 2016-02-04
Fixes #13451 - enables lazy sync

cfouant:


ehelms:
Does Pulp turn this on by default as well?

parthaa:
> Does Pulp turn this on by default as well?

Its off by default in pulp. We need it on for katello though.

manifests/params.pp
```diff
@@ -67,7 +67,7 @@
   $email_from = "no-reply@${::domain}"
   $email_enabled = false
 
-  $lazy_enabled = false
+  $lazy_enabled = true
```
ehelms: If Pulp doesn;t enable this by default, neither should the module. This should be up to users of the module such as puppet-katello to enable.

manifests/config.pp
```diff
@@ -97,4 +97,22 @@
       refreshonly => true,
     }
   }
+
+  if $pulp::lazy_enabled {
+    class { '::squid3':
```
ehelms: Is squid the Pulp supported option for lazy sync? @cristifalcas would you need a parameter here such as `manage_squid` or is this setup and configuration of squid if lazy sync is enabled OK?

jlsherrill: its not the only supported option, any http proxy would work i would imagine, but they provide same configs for two different proxies, squid and one more that isn't in the base OS.

jlsherrill: also pulp removed the lazy_enabled option from their config in the latest 2.8 nightly:  https://github.com/pulp/pulp/commit/0f100dbb81db860753cc97958bc315bc57eee4bc

so maybe it makes more sense to just move to to a manage_squid option

manifests/params.pp
```diff
@@ -67,7 +67,7 @@
   $email_from = "no-reply@${::domain}"
   $email_enabled = false
 
-  $lazy_enabled = false
+  $manage_squid = true
```
jlsherrill: i think we want this to be false by default, and turn it on here: 
https://github.com/Katello/puppet-katello/blob/master/manifests/init.pp#L96-L119
and 
https://github.com/Katello/puppet-capsule/blob/master/manifests/init.pp#L163-L187
and
https://github.com/Katello/puppet-katello_devel/blob/master/manifests/init.pp#L126-L144

and yes there has been discussion on consolidating these parts :)

templates/server.conf.erb
```diff
@@ -408,8 +408,8 @@ enabled: <%= scope['pulp::email_enabled'] %>
 #   downloading content from the Squid cache.
 
 [lazy]
-enabled: <%= scope['pulp::lazy_enabled'] %>
-<% if scope['pulp::lazy_enabled'] -%>
+enabled: <%= scope['pulp::manage_squid'] %>
+<% if scope['pulp::manage_squid'] -%>
```
jlsherrill: you can delete these two lines (and the <% end -%> below it.  This option doesn't exist anymore (hence the removal of the puppet param for lazy_enabled)

manifests/params.pp
ehelms: If lazy is simply just "on" in Pulp now, and it requires a proxy like this squid setup to be configured then having this set to true by default is fine. I equate this to how it manages and configures mongo and apache by default.

jlsherrill: Its not that its 'on' or 'off'.  Its more that if you create a repo with a download policy of 'on_demand' or 'background' it requires squid to be configured and running.  

I believe pulp is still using the 'immediate' policy by default which means normal pulp users not using the new feature won't need squid

ehelms: Does it just continue to work as before if you use those settings but do not have squid?

jlsherrill: it does, yes

jlsherrill: if you use the immediate setting i mean, if you use on_demand or background, requests from clients will get a 404 i believe

ehelms: Seems like a bad user experience, but whatever we can go this route and flip the bit if a user of the module comes along and wants it flipped.

jlsherrill: works4me

manifests/config.pp
```diff
@@ -97,4 +97,27 @@
...
+      $deprecated_opts = false
+    }
+    class { '::squid3':
+      use_deprecated_opts           => $deprecated_opts,
+      http_port                     => [ '3128 accel' ],
+      acl                           => [ 'Safe_ports port 3128' ],
+      http_access                   => [ 'allow localhost', 'deny to_localhost', 'deny all' ],
+      cache                         => [ 'allow all' ],
+      maximum_object_size           => '5 GB',
+      maximum_object_size_in_memory => '100 MB',
```
jlsherrill: I might would make this smaller, I'm thinking somewhere closer to 1MB.  I'm okay leaving this as is for now and adjusting later, but I think it needs to be tested for memory consumption at some point.

ehelms:
ACK - thanks @cfouant !

PR #124, opened 2016-02-09
Fixes #13607 - Adding pulp.conf

parthaa:
Moved the creation of pulp.conf to puppet-pulp instead of
katello/katello_devel/capsule.

Pulp makes a better spot than any of those.


manifests/config.pp
```diff
@@ -21,6 +21,16 @@
     target => $pulp::ca_cert,
   }
 
+  unless $pulp::manage_httpd {
```
ekohl: This sounds like reversed logic.

manifests/config.pp
```diff
@@ -21,6 +21,16 @@
...
+    file {'/etc/httpd/conf.d/pulp.conf':
```
ekohl: I wonder if this should be scoped to a vhost.

templates/pulp.conf.erb
```diff
@@ -0,0 +1,59 @@
...
+# DEBUG - uncomment the next 2 lines to enable debugging
+#WSGIRestrictStdin Off
+#WSGIRestrictStdout Off
+
+WSGISocketPrefix run/wsgi
+WSGIScriptAlias /pulp/api /usr/share/pulp/wsgi/webservices.wsgi
+WSGIImportScript /usr/share/pulp/wsgi/webservices.wsgi process-group=pulp application-group=pulp
+
+<Directory /usr/share/pulp/wsgi>
+<%- if scope.function_versioncmp([@apache_version, '2.4']) >= 0 -%>
```
ekohl: For this you need to pull in the variable in config.pp like https://github.com/Katello/puppet-capsule/blob/8d221b918730cb312073b2e7c85a36e2852d4252/manifests/init.pp#L137-L138 does.

ekohl:
I fully agree this is the proper module to set up pulp + apache.

manifests/config.pp
```diff
@@ -21,6 +21,16 @@
...
+      ensure  => file,
+      content => template('katello_devel/pulp.conf.erb'),
```
jlsherrill: katello_devel?

parthaa: AHH!

manifests/config.pp
parthaa: Intentional. Apparently 'manage_httpd' implies we want pulp module to configure httpd/vhost for us. Look at https://github.com/Katello/puppet-pulp/blob/master/manifests/apache.pp#L8 where it sets up the vhost. This is meant for stand alone installs, as in machines where only pulp runs.

However in the case of Katello we do not want to setup such a vhost. infact manage_httpd is set to false (https://github.com/Katello/puppet-katello/blob/master/manifests/init.pp#L114). We need pulp.conf to be setup in that scenario. I suggested creating a new var $manage_pulp_conf but @ehelms asked me to use manage_httpd instead...

manifests/config.pp
parthaa: see above comment

ehelms: This should probably just go in apache.pp and then you can use and `else` to deploy this instead of sticking it here with an unless.

manifests/config.pp
ekohl: That complicates matters. We use the apache_version from the apache module. Won't that module try to manage apache and interfere? At the very least I think a short comment describing why it is intentional is in order.

manifests/config.pp
ekohl: That makes a lot of sense.

templates/pulp.conf.erb
parthaa: Thanks!. Added.

manifests/apache.pp
```diff
@@ -5,6 +5,7 @@
   include ::apache::mod::ssl
   include ::apache::mod::xsendfile
 
+  $apache_version = $::apache::apache_version
```
ehelms: A blank line after this would be nice for readability.

templates/pulp.conf.erb
```diff
@@ -0,0 +1,59 @@
...
+#
+
+# Apache configuration file for pulp web services and repositories
+
+AddType application/x-pkcs7-crl .crl
+AddType application/x-x509-ca-cert .crt
+
+# Example ssl cert and key files to get you started.
+# This MUST match /etc/pulp/pulp.conf [security] 'cacert'.
+SSLCACertificateFile <%= scope.lookupvar("::pulp::ca_cert") %>
```
ehelms: Nitpick: prefer `scope['::pulp::ca_cert']` syntax here

ekohl:
untested, but ACK code wise

manifests/apache.pp
parthaa: added.

ehelms:
ACK - thanks @parthaa !

PR #126, opened 2016-02-10
do not specify ca cert in apache pulp.conf

jlsherrill:


beav:
@jlsherrill  typo in commit msg

jlsherrill:
@beav hah good catch!

beav:
is this line needed to validate that entitlement certs are signed when making requests for content?

jlsherrill:
@beav which line exactly?

jlsherrill:
@beav oh!  It should be specified by the virtual host controlling 443 as well

beav:
@jlsherrill i see it now. Thanks!!

ACK from me

PR #128, opened 2016-02-18
Add KeepAlive support

jlsherrill:


templates/etc/httpd/conf.d/_ssl_vhost.conf.erb
```diff
@@ -0,0 +1,7 @@
+# allow older yum clients to connect, see bz 647828
+SSLInsecureRenegotiation On
```
beav: :+1: for keeping this around, given older EL5/Centos5 clients need to connect in order to update

beav:
@jlsherrill for my own understanding, this PR adds the keepalive and does a bit of cleanup to remove `_ssl_username.conf.erb`? just confirming whats change vs what's refactor

manifests/init.pp
```diff
@@ -247,6 +247,9 @@
 # $node_oauth_key::             The oauth key used to authenticate to the parent node
 #
 # $node_oauth_secret::          The oauth secret used to authenticate to the parent node
+#
+# $max_keep_alive::             Configuration value for apache MaxKeepAliveRequests
```
ekohl: Add ```type:integer``` for consistency?

jlsherrill:
@beav correct, i wanted to make sure this was there if the user has manage_apache set to true or not, and saw there was a good reason to do a small refactor to use a single fragment.

beav:
@jlsherrill I ran this on a bats machine but did not see any update to add KeepAlive. Does this test look valid?

 * run bats
 * replace `/usr/share/katello-installer/modules/pulp` with this checkout and branch
 * re-run katello-installer
 * look for KeepAlive option in `/etc/httpd/conf.d`

jlsherrill:
@beav as far as pulp goes this change only affects you if you a) are using manage_apache = true on a main pulp server or b) are using a 'child' node.    So this change only really affects katello on the capsule.  My PR here:  https://github.com/Katello/puppet-katello/pull/110  does add this for the main katello installation

beav:
Patch works as advertised. ACK from me after https://github.com/Katello/puppet-pulp/pull/128#discussion_r53468284 is addressed

manifests/init.pp
jlsherrill: added!

PR #130, opened 2016-02-22
adds pulp streamer service

cfouant:


manifests/service.pp
```diff
@@ -34,4 +34,12 @@
     status     => 'service pulp_resource_manager status | grep "node resource_manager"',
   }
 
+  service { 'pulp_streamer':
+    ensure     => running,
+    require    => [Service[mongodb]],
```
cristifalcas: I'm not on the pulp team, but I would like to request to not force the pulp server to be on the same machine as the mongodb.

jlsherrill: @cfouant should be able to check $manage_db and only require that if true.

manifests/apache.pp
```diff
@@ -80,6 +80,16 @@
     }
   }
 
+  if $pulp::manage_db {
```
ehelms: Why manage_db?

manifests/apache.pp
```diff
@@ -80,6 +80,16 @@
...
+    file {'/etc/pulp/streamer.conf':
```
ehelms: I think this should go in config.pp

ehelms: Actually, guessing this path is just wrong and you meant it to go in `/etc/httpd/conf.d` based upon the streamer.conf template

daviddavis: @ehelms is correct. this should be `/etc/httpd/conf.d/pulp_streamer.conf`.

manifests/service.pp
jlsherrill: @cristifalcas after some discussion, we do agree that the mongodb service should not be required for this service, its already required for all the other services so we should probably just address them all together.

daviddavis: Added a redmine issue: http://projects.theforeman.org/issues/13864

daviddavis:
Tested and it worked. ACK from me.

manifests/service.pp
ehelms: Looks like we can add this to (https://github.com/Katello/puppet-pulp/blob/master/manifests/database.pp#L14-L17) and remove these requires. I agree lets do a follow up with that change.

ehelms:
Thanks @cfouant !

PR #131, opened 2016-02-25
make pulp streamer optional

cerealcake:
make pulp_streamer installation optional and include apache::mod::proxy in the event that pulp_streamer is included with the apache config added in https://github.com/Katello/puppet-pulp/commit/3e9195baca54a50745b591e16ac8ff501134d26a

### DIFF
--- a/manifests/apache.pp
+++ b/manifests/apache.pp
@@ -117,11 +117,14 @@ class pulp::apache {
     }
   }
 
-  file {'/etc/httpd/conf.d/pulp_streamer.conf':
-    ensure  => file,
-    content => template('pulp/etc/httpd/conf.d/pulp_streamer.conf.erb'),
-    owner   => 'root',
-    group   => 'root',
-    mode    => '0644',
+  if $pulp::enable_streamer {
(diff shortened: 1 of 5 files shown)


ehelms:
@cfouant @daviddavis can you double check this still meets your needs?

ekohl:
The change itself looks good. I don't see any changes to the vhost configuration but I don't think that includes any streamer config anyway.

cerealcake:
Noticed this is only required if not using a version greater or equal to Pulp-2.8.0, as pulp-streamer is available with this version.

I believe this can be closed as https://repos.fedorapeople.org/pulp/pulp/stable/ now includes Pulp-2.8.

ekohl:
The ```include ::apache::mod::proxy``` is something we should add.

stbenjam:
Needs rebase

cerealcake:
As per my comment above https://repos.fedorapeople.org/pulp/pulp/stable/ now includes Pulp-2.8. which includes pulp-streamer.

For an item which is available on stable, it may not require to be optional anymore.

PR #132, opened 2016-02-29
update vhosts80/rpm.conf to support http lazy sync

jlsherrill:


jlsherrill:
Note this matches what is now shipped with pulp  https://github.com/pulp/pulp_rpm/blob/master/plugins/etc/pulp/vhosts80/rpm.conf

dLobatog:
Can confirm it works. With that line on, requests such as http://katello/pulp/repos/Lobacorp/Library/content/dist/rhel/server/7/7.2/x86_64/kickstart/images/pxeboot/initrd.img would simply 403 and are not symlinked to anywhere

templates/vhosts80/rpm.conf
```diff
@@ -1,3 +1,2 @@
-Alias /pulp/repos /var/www/pub/yum/http/repos
```
ehelms: So these aliases are no longer needed for RPM content; lazy handles it differently?

jlsherrill: This one alias is no longer needed.  Instead the streamer wsgi app handles it and decides whether to serve the content or fetch it from upstream.

ehelms:
ACK

PR #133, opened 2016-03-10
Allow specifying custom fragment on httpd module

ehelms:


manifests/apache/fragment.pp
```diff
@@ -0,0 +1,20 @@
...
+#
+#  === Parameters:
+#
+#  $ssl_content:: content of the ssl virtual host fragment
+define pulp::apache::fragment(
+  $ssl_content = undef,
+  $order       = 15,
+) {
+
+  if $ssl_content {
```
ekohl: Does it make sense to use this fragment if there is no content?

spec/defines/pulp_apache_fragment_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+      }
+    end
+
+    context 'with ssl_content parameter' do
+      let :params do
+        { :ssl_content => "some_string" }
+      end
+
+      it do
+        should contain_concat__fragment("fragment_title").with_content(/some_string/)
```
ekohl: You can make this explicit with quotes instead of a regex.

spec/defines/pulp_apache_fragment_spec.rb
```diff
@@ -0,0 +1,34 @@
+require 'spec_helper'
+
+describe 'pulp::apache::fragment' do
+  let(:title) { 'fragment_title' }
+
+  let :pre_condition do
```
ekohl: Is this needed for the concat target?

manifests/apache/fragment.pp
ehelms: My brain pre-optimized to allow HTTP  fragments in the future, I can just remove this check for now :)

ekohl: Then you can also remove the ```= undef``` and add a validate_string.

johnpmitsch:
ACK tested with https://github.com/Katello/puppet-capsule/pull/76 and https://github.com/Katello/puppet-certs/pull/78 on a capsule install and install ran successfully

PR #136, opened 2016-03-17
update to add defaultsite to squid conf

beav:
Squid 3.1 (el6/centos6 default version) requires this directive to be set for
the Pulp streamer. Without it, the streamer will return 400 errors.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -106,7 +106,7 @@ class pulp::config {
     }
     class { '::squid3':
       use_deprecated_opts           => $deprecated_opts,
-      http_port                     => [ '3128 accel' ],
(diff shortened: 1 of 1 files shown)


jlsherrill:
i assume this works on el7 as well?

beav:
@jlsherrill it should but testing now. el7 uses squid 3.2, i assume the change is compat with both

beav:
@jlsherrill success on both 6 and 7

test was:

* bats install nightly
* make changes to config.pp
* re-run foreman-installer, confirm changes landed
* run bats content test

jlsherrill:
ACK thanks @beav !

PR #140, opened 2016-04-16
Add paths for puppet-lint docs check

stbenjam:
This checks that each of the class parameters is documented. We need this check
on classes exposed in the installer directly.

### DIFF
--- a/.sync.yml
+++ b/.sync.yml
@@ -1,0 +1,6 @@
+---
+Rakefile:
+  param_docs_pattern:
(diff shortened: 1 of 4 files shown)


ekohl:
Looks like some parameters aren't documented.

stbenjam:
Indeed, documented now

ekohl:
ACK

stbenjam:
Thanks!

PR #144, opened 2016-05-03
allow browsing root of repos

jlambert121:


ehelms:
Thanks @jlambert121 !

PR #145, opened 2016-05-11
fixes #15014 - restore pulp db init flag

stbenjam:
This brings back the init flag so that the puppet installer only
runs the migration once or when the user has run --reset (which
deletes the init.flag file)

### DIFF
--- a/manifests/database.pp
+++ b/manifests/database.pp
@@ -18,12 +18,12 @@ class pulp::database {
   }
 
   exec { 'migrate_pulp_db':
-    command     => 'pulp-manage-db',
-    path        => '/bin:/usr/bin',
-    logoutput   => 'on_failure',
-    user        => 'apache',
-    refreshonly => true,
-    require     => File['/etc/pulp/server.conf'],
+    command   => 'pulp-manage-db && touch /var/lib/pulp/init.flag',
+    path      => '/bin:/usr/bin',
+    logoutput => 'on_failure',
+    user      => 'apache',
+    creates   => '/var/lib/pulp/init.flag',
+    require   => File['/etc/pulp/server.conf'],
   }
 
   Class['pulp::install'] ~> Exec['migrate_pulp_db'] ~> Class['pulp::service']


ekohl:
ACK. I recall there was an issue where someone was complaining the module wasn't idempotent because mongodb was restarted on every run. I can't find it now, but it looks like this should fix it.

stbenjam:
Hm, not sure if this will fix that, this will cause a migrate whenever the service was down and puppet started it. Since the yum upgrade could happen out of band, this is the only way I could think of to ensure the database is  up-to-date when we run it.

ehelms:
On the surface, this seems like a strange paradigm, every time the database is started migrate it. Does the foreman module for example follow this paradigm?

stbenjam:
> On the surface, this seems like a strange paradigm, every time the database is started migrate it. Does the foreman module for example follow this paradigm?

It is a very strange to do, yes.  Do note, it's not every time the database is started, only when puppet does it.

The problem with `--reset` is the mongodb is never getting migrated.  So, my idea was to leave it off in the reset pre-hook, so puppet starts it and triggers the migrate.  

Do you think there's some other thing we could subscribe to that indicates the DB needs migrating? Maybe re-introduce `/var/lib/pulp/init.flag` which seems to have been removed?

stbenjam:
(init.flag and the reset stuff was removed in https://github.com/Katello/puppet-pulp/pull/65)

beav:
I got bit by this bug yesterday, :+1: from me for this patch. Running `pulp-manage-db` is pretty quick, especially after a `--reset` :koala:

ehelms:
If @cristifalcas can tell us why he removed it (whether by accident or with an intent) we could maybe bring back the init file otherwise the only thing I could think of would be tying the migrate to the creation of the dbpath so that it migrates whenever that is refreshed/recreated. Failing any of that I'll be OK with this change.

cristifalcas:
Hi, from what I remember, I did it because I didn't see any purpose for it. Also it was stopping future migrations if nobody removed the file. It may be useful for tests, but in production not so much.

stbenjam:
>  I did it because I didn't see any purpose for it.

It's used in production by Katello, the `--reset` command to the installer removes the file.   Perhaps the database path for mongo makes more sense than the init flag file

> Also it was stopping future migrations if nobody removed the file. 

This will never happen now anyway, a change on a `require` doesn't trigger refresh on puppet resource, AFAIK.

manifests/database.pp
```diff
@@ -23,7 +23,7 @@
     logoutput   => 'on_failure',
     user        => 'apache',
     refreshonly => true,
-    require     => File['/etc/pulp/server.conf'],
```
cristifalcas: I think this is still need it because the migration should run after all needed information is populated.

stbenjam: I suppose, is there a reason this is an include, and not between config + service?

 https://github.com/Katello/puppet-pulp/blob/master/manifests/init.pp#L377

stbenjam:
If there's no better suggestions, I think I'll go with just putting the init flag back.

stbenjam:
Ready for another look

jlsherrill:
ACK, tested and worked for me.

stbenjam:
Thanks!

PR #147, opened 2016-05-17
refs #15058 - mongo auth only on newer mongos

stbenjam:
puppetlabs-mongodb doesn't manage users correctly on < 2.6.

### DIFF
--- a/manifests/database.pp
+++ b/manifests/database.pp
@@ -7,7 +7,12 @@ class pulp::database {
       $mongodb_pidfilepath = '/var/run/mongodb/mongodb.pid'
     }
 
-    $auth_real = $::pulp::db_username != undef and $::pulp::db_password != undef
+    # puppetlabs-mongodb is totally broken with managing users on < 2.6
+    if (versioncmp($::mongodb_version, '2.6.0') >= 0) {
+      $auth_real = $::pulp::db_username != undef and $::pulp::db_password != undef
+    } else {
+      $auth_real = false
+    }
 
     class { '::mongodb::server':
       pidfilepath => $mongodb_pidfilepath,
--- a/templates/server.conf.erb
+++ b/templates/server.conf.erb
@@ -47,11 +47,15 @@
 [database]
 name: <%= scope['pulp::db_name'] %>
 seeds: <%= scope['pulp::db_seeds'] %>
+<%# Auth is currently only working on mongo >= 2.6 -%>
+<% mongo_version = scope.lookupvar('::mongodb_version') -%>
+<% if mongo_version && scope.function_versioncmp([mongo_version, '2.6.0']) >= 0 -%>
 <% unless [nil, :undefined, :undef, ''].include?(scope.lookupvar('pulp::db_username')) -%>
 username: <%= scope['pulp::db_username'] %>
 <% end -%>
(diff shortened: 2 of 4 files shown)


beav:
the WIP looks reasonable to me

stbenjam:
Oh, it'll be more complicated than this because we need to turn off auth in pulp server.conf too

stbenjam:
Get a clean install on el6 with these changes. This seems to work for me, now the question is if we want to do it this way.  It means db_username/db_password are unused on el6 until puppetlabs-mongodb is fixed, which may be never.  

There's an open PR at https://github.com/puppetlabs/puppetlabs-mongodb/pull/253, but it looks abandoned.

ekohl:
There was a discussion about deprecating EL6 starting with Foreman 1.12 so I'm fine with limited functionality on EL6. We should mention it in the release notes though.

manifests/database.pp
```diff
@@ -7,26 +7,31 @@
...
+      $auth_real = false
+    }
 
     class { '::mongodb::server':
       pidfilepath => $mongodb_pidfilepath,
       auth        => $auth_real,
       noauth      => !$auth_real,
     }
 
-    if $auth_real {
```
ekohl: No need to move this block if ```$auth_real``` is always false on older versions.

stbenjam: It looked better to me to have all the logic in one block, so you know why we're doing this, but I can move it back if you'd like

stbenjam: Nevermind, it looks better with your suggestion

templates/server.conf.erb
```diff
@@ -47,12 +47,16 @@
 [database]
 name: <%= scope['pulp::db_name'] %>
 seeds: <%= scope['pulp::db_seeds'] %>
+<%# Auth is currently only working on mongo > 2.4 -%>
+<% mongo_version = scope.lookupvar('::mongodb_version') -%>
+<% if mongo_version && mongo_version.split('.')[0] == '2' && mongo_version.split('.')[1].to_i > 4 -%>
```
ekohl: Can't you use the versioncmp function inside a template?

stbenjam: Oh, you can!

templates/server.conf.erb
```diff
@@ -47,12 +47,16 @@
 [database]
 name: <%= scope['pulp::db_name'] %>
 seeds: <%= scope['pulp::db_seeds'] %>
+<%# Auth is currently only working on mongo > 2.4 -%>
```
ekohl: It can be confusing if you use > 2.4 here and >= 2.6 below.

templates/server.conf.erb
```diff
@@ -47,12 +47,16 @@
...
+<% mongo_version = scope.lookupvar('::mongodb_version') -%>
+<% if mongo_version && scope.function_versioncmp([mongo_version, '2.6']) >= 0 -%>
```
ekohl: nit: you use 2.6 here and 2.6.0 in the puppet code. I'd use the exact same version string.

ekohl:
:+1: these comments are just minor nits.

ekohl:
ACK on green tests

stbenjam:
Thanks @ekohl, merging

PR #149, opened 2016-05-31
refs #15326 - remove mongo authentication

stbenjam:
It is broken on RHEL 6, and does not seem to function reliably on
RHEL 7.  Subsequent runs of the installer will generate errors.

Because of https://tickets.puppetlabs.com/browse/MODULES-534, it  seems, which remains unfixed for quite some time.  It seems clear to me know *why* auth is false by default in this thing, the module is really flaky.  

Since we're planning to drop mongo in a future release anyway, I think it's best just to leave it off on el6 and el7 now too.

Any thoughts on this?

### DIFF
--- a/manifests/database.pp
+++ b/manifests/database.pp
@@ -7,29 +7,8 @@ class pulp::database {
       $mongodb_pidfilepath = '/var/run/mongodb/mongodb.pid'
     }
 
-    # puppetlabs-mongodb is totally broken with managing users on < 2.6
-    if (versioncmp($::mongodb_version, '2.6.0') >= 0) {
-      $auth_real = $::pulp::db_username != undef and $::pulp::db_password != undef
-    } else {
-      $auth_real = false
-    }
-
     class { '::mongodb::server':
       pidfilepath => $mongodb_pidfilepath,
-      auth        => $auth_real,
-      noauth      => !$auth_real,
-    }
-
-    if $auth_real {
-      mongodb_user { $::pulp::db_username:
-        ensure        => present,
-        username      => $::pulp::db_username,
-        password_hash => mongodb_password($::pulp::db_username, $::pulp::db_password),
-        database      => $::pulp::db_name,
-        roles         => ['dbAdmin', 'readWrite'],
-        tries         => 10,
-        require       => Service['mongodb'],
-      }
     }
 
     Service['mongodb'] -> Service['pulp_celerybeat']


stbenjam:
Not sure what to do about the params we added to puppet-katello/puppet-capsule/puppet-katello_devel either.

Removing them probably means a version bump, right?

ekohl:
:+1: for dropping if it's unreliable. Especially when it's going to be dropped in the near future.

> Removing them probably means a version bump, right?

Only if it was in a released version.

stbenjam:
After thinking about it, I think it makes sense to leave all the db_username/db_password stuff in puppet-pulp at least. If someone setup an external mongo manually then we'd still support authenticating to it, we just can't/won't create the user if it doesn't already exist.  Does that make sense?

stbenjam:
See also:

- https://github.com/Katello/puppet-katello/pull/130
- https://github.com/Katello/puppet-capsule/pull/85
- https://github.com/Katello/puppet-katello_devel/pull/79

ehelms:
Is there a way to support user creation via an exec or custom type to get around whatever the mongodb module is doing wrong?

akofink:
LGTM, though I'm not super familiar with this module.

stbenjam:
> Is there a way to support user creation via an exec or custom type to get around whatever the mongodb module is doing wrong?

It'd be a huge mess, as it's multiple commands requiring json input into mongo CLI. You can see how puppet provider does it:

https://github.com/puppetlabs/puppetlabs-mongodb/blob/master/lib/puppet/provider/mongodb_user/mongodb.rb

And it doesn't even work after the initial user creation.  The workflow is something like this:

1. Execute command to make sure user doesn't exist yet
2. Create the mongo-formatted password hash
3. Execute the command to create the user
4. Execute the command to update the password (since it could be changed in puppet param)
5. Add the required roles to the user

The problem is then once you actually turn on the auth, puppet can't run any of those commands, because it now needs an admin user to authenticate with! So you'll then need check if auth is on, and if it is, then login before doing all of the above commands.

And of course the JSON you need to supply on 2.4 and 2.6 is different.

It's really a clusterf!@#$.

stbenjam:
Which is why I was inclined to just (╯°□°）╯︵ ┻━┻ and wait until we use a sensible database like PostgreSQL.  If we really need this, I could try to fix the puppet provider but my experience on even getting one-liners merged into that repo is it takes ages for them to act on them (the PR to fix 2.4 has been open for months and months).

ehelms:
ACK

Issue #150, opened 2016-06-07
Catalogue failure if manage_db=false


ccnifo:
Services defined in service.pp have requires on _Service[mongodb]_. This is blocking if you have set "manage_db = false" (for example, your DB server is on another node): actually, there's no resource _Service[mongodb]_ in this case, which is normal but leads to catalogue failure.

First of all: aren't the first 3 require in service.pp completely redundant with those in database.pp?
```
    Service['mongodb'] -> Service['pulp_celerybeat']
    Service['mongodb'] -> Service['pulp_workers']
    Service['mongodb'] -> Service['pulp_resource_manager']
```
It would thus be possible to remove them from service.pp, without forgetting to add     `Service['mongodb'] -> Service['pulp_streamer']` in database.pp.

Else, we may write a create_resource for the 4 services, with a switch related to the manage_db value.

PR #153, opened 2016-06-23
fixes GH-150 - Eliminate duplicate, broken requires on mongodb

ccnifo:
This is a fix for https://github.com/Katello/puppet-pulp/issues/150: it deletes duplicate requires and, more important, let them be included in the catalogue only if the resource "service['mongodb']" has been created.

This change should make the module work for people using "manage_db=false" (e.g. MongoDB installed on another node).


ekohl:
ACK on green tests

stbenjam:
Thanks @ccnifo 

ACK from me too

PR #155, opened 2016-07-26
allow configuration of puppet wsgi processes

jlsherrill:


manifests/params.pp
```diff
@@ -99,6 +99,8 @@
   $max_keep_alive = 10000
   $num_workers = min($::processorcount, 8)
 
+  $puppet_wsgi_processes = 3
```
ekohl: Any particular reason to pick 3 here?

jlsherrill: its the default value  pulp provides:  https://github.com/pulp/pulp_puppet/blob/master/pulp_puppet_plugins/etc/httpd/conf.d/pulp_puppet.conf#L39

manifests/init.pp
```diff
@@ -254,6 +254,8 @@
 # $max_keep_alive::             Configuration value for apache MaxKeepAliveRequests
 #                               type:integer
 #
+# $puppet_wsgi_processes::      Number of WSGI processes to spawn for the puppet webapp
```
ekohl: It needs to be included in the header below as well

jlsherrill: doh!  thanks, updated

ekohl:
ACK on green tests

manifests/params.pp
ekohl: I noticed that after I made the comment. That's fine then, if needed it can be tuned later like num_workers above.

jlsherrill:
thanks @ekohl !

PR #156, opened 2016-08-04
Do not attempt to write `vhosts80` file too early

beav:
Puppet was attempting to create `/etc/pulp/vhosts80` before `pulp-server` was
installed.  This led to errors because `/etc/pulp` didn't exist yet.


ehelms:
ACK - thanks @beav

PR #158, opened 2016-08-26
Fixes #16343 - Redirect fix for Atomic Hosts

parthaa:
Rhel 7 Docker Containers inside an Atomic host need to be able to pull
rpms from within /pulp/ostree/web/. But those are available only under
/pulp/repos

This commit sets up a config to redirect 'yum repo like' calls to
/pulp/repos to facilitate that

### DIFF
--- a/spec/classes/pulp_apache_spec.rb
+++ b/spec/classes/pulp_apache_spec.rb
@@ -311,7 +311,6 @@ Alias /pulp/python /var/www/pub/python/
       end
     end
 
-
     describe 'with enable_ostree' do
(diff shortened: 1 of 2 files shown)


ekohl:
I'd prefer to use [$rewrites on the vhost](https://github.com/puppetlabs/puppetlabs-apache#rewrites)

ekohl:
Never mind that, it's very hard to get that in given the repo structure.

ehelms:
ACK - thanks @parthaa !

PR #159, opened 2016-08-29
this fact make a useless warn message

PascalBourdier:
rpmquery is provide by rpm but repoquery is in yum-utils

we have this message on several agents :
sh: repoquery: command not found


ekohl:
https://github.com/Katello/puppet-pulp/pull/115 should address this in a more generic way so closing this as duplicate.

PR #162, opened 2016-08-30
Fix typo in variable name in params.pp

stbenjam:


ehelms:
Blast, seems to be more failing now.

stbenjam:
`root_home` is a fact from stdlib, @ekohl How do you handle these? They're not included in the os_facts hashes, as they're from puppetlabs-stdlib.  Should we just add them to the hash ourselves? Seems less than ideal, what happens next time puppetlabs-mongodb uses a stdlib - maybe we could come up with a better solution?

ekohl:
@stbenjam we do the same with the concat tmpdir so I think we should do the same here.

ehelms:
Looks like the Puppet 4 jobs failed needing this to be updated too:

https://github.com/Katello/puppet-pulp/blob/master/spec/defines/pulp_child_fragment_spec.rb#L15

ehelms:
ACK

stbenjam:
THANKS!

PR #165, opened 2016-09-13
ssl_protocol: customisable parameter and default avoids SSLv3

ccnifo:
The SSLProtocol directive is hardcoded as "_all -SSLv2_" in the _pulp-https_ vhost ressource declaration. Puppetlabs/apache would have put it to "_all -SSLv2 -SSLv3_" by default, because SSLv3 is now deprecated (RFC 7568).
This PR makes _ssl_protocol_ a customisable parameter and sets its default value to the current apache module default.

### DIFF
--- a/manifests/apache.pp
+++ b/manifests/apache.pp
@@ -32,7 +32,7 @@ class pulp::apache {
       ssl_chain                  => $pulp::https_chain,
       ssl_ca                     => $pulp::ca_cert,
       ssl_verify_client          => 'optional',
-      ssl_protocol               => ' all -SSLv2',
+      ssl_protocol               => $pulp::ssl_protocol,
       ssl_options                => '+StdEnvVars +ExportCertData',
       ssl_verify_depth           => '3',
       wsgi_process_group         => 'pulp',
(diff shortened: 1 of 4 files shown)


ekohl:
The puppet code mostly looks good, but you can add some validation:

```puppet
if $ssl_protocol != undef {
  validate_string($ssl_protocol)
}
```

I'm not familiar enough with TLS support in older RHELs to know if this would cause problems for some clients. @stbenjam / @ehelms do you know by any chance?

ccnifo:
Thank you for the comment, validation added.

About TLS support, RH knowledge base states this : 
> Support for TLS by Red Hat Enterprise Linux version:
>
>Red Hat Enterprise Linux 5: SSL V3 and TLS 1.0
>Red Hat Enterprise Linux 6: SSL V3, TLS 1.0, TLS 1.1 and TLS 1.2
>Red Hat Enterprise Linux 7: SSL V3, TLS 1.0, TLS 1.1 and TLS 1.2

and discourages SSLv3 usage+recommends TLS 1.2 whenever possible.

I didn't find any info about RHEL4.

ekohl:
Personally I'm fine with the new defaults because I don't care about RHEL4 but I'd like an ACK from other devs on this.

ehelms:
Looks good to me, thanks @ccnifo !

PR #166, opened 2016-09-14
Restrict SSLUsername to /pulp/api

jlsherrill:
otherwise apache will warn about any request that does
not use a client cert


ehelms:
Remind me -- what we are using to allow Pulp to sync HTTPs repositories from the server when it's on a Capsule?

jlsherrill:
Katello ------>  capsule pulp api  uses the pulp client cert (that SSLUsername depends on)
Katello <------ Capsule (fetching content) uses the ueber cert

ehelms:
ACK

PR #167, opened 2016-09-20
allow to increase the timeout for the exec that does the db migration

cristifalcas:
On our machines the exec  'migrate_pulp_db' times out during pulp upgrades. To fix this we have to run the command manually on all servers.

This parameter should fix this issue.


ehelms:
Thanks @cristifalcas !

PR #170, opened 2016-11-02
Move crane to puppet-pulp

stbenjam:
Crane is a pulp project, and should be configured by puppet-pulp.

As part of decommissioning puppet-capsule, it would be easier
if we could decommission puppet-crane too and move puppet-pulp
to the top-level of the installer.


stbenjam:
I think it should be configurable from the the main pulp class, would it make sense to have an `$enable_crane` parameter?

manifests/crane.pp
```diff
@@ -0,0 +1,34 @@
...
+  $data_dir = $pulp::crane::params::data_dir,
+  ) inherits pulp::crane::params {
+
+  validate_absolute_path($key, $cert, $ca_cert, $data_dir)
+  validate_integer($port)
+
+  class { '::pulp::crane::install': } ~>
+  class { '::pulp::crane::apache': } ~>
+  class { '::pulp::crane::config': } ~>
+  Service['httpd'] ->
```
ekohl: I think this notify is redundant. If ```apache::vhost``` doesn't do this, something is seriously wrong.

manifests/crane.pp
```diff
@@ -0,0 +1,34 @@
...
+  $port     = $pulp::crane::params::port,
+  $data_dir = $pulp::crane::params::data_dir,
+  ) inherits pulp::crane::params {
+
+  validate_absolute_path($key, $cert, $ca_cert, $data_dir)
+  validate_integer($port)
+
+  class { '::pulp::crane::install': } ~>
+  class { '::pulp::crane::apache': } ~>
+  class { '::pulp::crane::config': } ~>
```
ekohl: Wouldn't you want the config before apache?

manifests/crane.pp
```diff
@@ -0,0 +1,34 @@
...
+  $cert     = $pulp::crane::params::cert,
+  $ca_cert  = $pulp::crane::params::ca_cert,
+  $port     = $pulp::crane::params::port,
+  $data_dir = $pulp::crane::params::data_dir,
+  ) inherits pulp::crane::params {
+
+  validate_absolute_path($key, $cert, $ca_cert, $data_dir)
+  validate_integer($port)
+
+  class { '::pulp::crane::install': } ~>
```
ekohl: I don't think the install needs to notify but I don't think it hurts either.

manifests/crane/params.pp
```diff
@@ -0,0 +1,8 @@
+# Default params for Crane
+class pulp::crane::params {
+  $fqdn = $::fqdn
```
ekohl: This is unused, but is a good idea. I'd think that ```$hostname``` is a better name though.

stbenjam:
Just FYI, this is an existing module that I'm combining with puppet-pulp, I didn't write any of this.

ekohl:
I know, but maybe we can clean it up while we're at it rather than move legacy around.

> I think it should be configurable from the the main pulp class, would it make sense to have an $enable_crane parameter?

I think that would make sense. I also wonder if we should reuse the certificate options already present in the module.

ehelms:
I guess this makes the most sense. Crane is technically a stand alone project that can work with Pulp but does not require Pulp as far as I understand it.

stbenjam:
> I guess this makes the most sense. Crane is technically a stand alone project that can work with Pulp but does not require Pulp as far as I understand it.

Indeed, and no requirement on a user to use the rest of this module, if they want crane, they can just use `pulp::crane`.

templates/crane.conf.erb
```diff
@@ -0,0 +1,9 @@
+# =========================
+# Pulp Crane Configuration
+# =========================
+
+[general]
+debug: false
+#data_dir_polling_interval: 60
```
ehelms: This is commented out and cannot be modified by a user, mind either removing it entirely or making it configurable?

manifests/crane/apache.pp
```diff
@@ -0,0 +1,28 @@
...
+                          {
+                            '/' => '/usr/share/crane/crane.wsgi',
+                          },
+    port                => $pulp::crane::port,
+    priority            => '03',
+    ssl                 => true,
+    ssl_cert            => $pulp::https_cert,
+    ssl_key             => $pulp::https_key,
+    ssl_chain           => $pulp::https_chain,
+    ssl_ca              => $pulp::ca_cert,
```
ehelms: Would this mean I have to configure Pulp to use `pulp::crane` ?

manifests/init.pp
```diff
@@ -266,6 +274,8 @@
 #
 class pulp (
   $version                   = $pulp::params::version,
+  $crane_port                = $pulp::params::crane_port,
+  $crane_data_dir            = $pulp::params::crane_data_dir,
```
stbenjam: I'm not sure, @ekohl could maybe tell us

manifests/crane/apache.pp
stbenjam: Yes, sorry, I'll fix it so you don't

manifests/init.pp
ekohl: It would mean that you have to include ```pulp::crane::params``` in ```pulp::params```. I'm not sure that's a good idea, mostly because it's static anyway.

manifests/crane.pp
```diff
@@ -0,0 +1,37 @@
...
+
+  validate_absolute_path($data_dir, $key, $cert, $ca_cert)
+  validate_integer($port, $data_dir_polling_interval)
+
+  class { '::pulp::crane::install': } ~>
+  class { '::pulp::crane::config': } ~>
+  class { '::pulp::crane::apache': } ->
+  Class['::pulp::crane']
+}
+
```
ekohl: Empty line

manifests/init.pp
```diff
@@ -394,6 +408,16 @@
   include ::pulp::database
   include ::pulp::broker
 
+  if($enable_crane) {
```
ekohl: The parenthesis aren't needed.

manifests/init.pp
```diff
@@ -376,6 +388,8 @@
   validate_array($disabled_authenticators)
   validate_hash($additional_wsgi_scripts)
   validate_integer($max_keep_alive)
+  validate_integer($crane_port)
```
ekohl: This is also validated the the crane class itself.

manifests/crane/apache.pp
```diff
@@ -0,0 +1,28 @@
+# Sets up Apache for Crane
+class pulp::crane::apache {
+
+  include ::apache
+  include ::apache::mod::headers
+  include ::apache::mod::proxy
+  include ::apache::mod::proxy_http
+
+  apache::vhost { 'crane':
+    servername          => $pulp::crane::fqdn,
```
ekohl: This variable currently doesn't exist.

manifests/crane/apache.pp
```diff
@@ -0,0 +1,28 @@
...
+    docroot             => '/usr/share/crane/',
+    wsgi_script_aliases =>
+                          {
+                            '/' => '/usr/share/crane/crane.wsgi',
+                          },
+    port                => $pulp::crane::port,
```
ekohl: This can be ```$::pulp::crane::port```. Same goes for the other parameters here.

manifests/crane.pp
```diff
@@ -0,0 +1,37 @@
...
+#
+# $port::                       Port for Crane to run on
+#                               type:integer
+#
+# $data_dir::                   Directory containing docker v1/v2 artifacts published by pulp
+#
+# $data_dir_polling_interval::  The number of seconds between checks for updates to metadata files in the data_dir
+#                               type:integer
+class pulp::crane (
+  $port                      = $pulp::crane::params::port,
```
ekohl: These can be ```$::pulp::crane::params::port```

manifests/crane.pp
```diff
@@ -0,0 +1,37 @@
...
+  $ca_cert                   = $pulp::crane::params::ca_cert,
+  ) inherits pulp::crane::params {
+
+  validate_absolute_path($data_dir, $key, $cert, $ca_cert)
+  validate_integer($port, $data_dir_polling_interval)
+
+  class { '::pulp::crane::install': } ~>
+  class { '::pulp::crane::config': } ~>
+  class { '::pulp::crane::apache': } ->
+  Class['::pulp::crane']
```
ekohl: I don't think this has to be absolute because evaluated classes are always absolute.

stbenjam:
@ekohl A couple of small changes, if you want to look again - I parameterized debug logging for crane

spec/classes/pulp_crane_spec.rb
```diff
@@ -0,0 +1,56 @@
+require 'spec_helper'
+
+describe 'pulp::crane' do
+
+ context 'on redhat' do
+    context 'with parameters' do
+      let :pre_condition do
```
ekohl: Since you're testing this class I believe you can use ```let :params``` instead of a pre condition.

stbenjam:
Thanks, changed to params. 

@ehelms Do you want to take a look before merging?

manifests/crane/params.pp
```diff
@@ -0,0 +1,10 @@
+# Default params for Crane
+class pulp::crane::params {
+  $ca_cert  = undef
+  $cert = undef
+  $data_dir = '/var/lib/crane/metadata'
```
ehelms: Minor question - why does this differ from the default in pulp/params.pp?

manifests/crane/params.pp
```diff
@@ -0,0 +1,10 @@
...
+  $data_dir_polling_interval = 60
+  $debug = false
+  $key = undef
```
ehelms: Nitpick - I realize these are in sorted order but the related items might be nice together.

ehelms:
Just the two small questions, otherwise looks great!

ehelms:
@stbenjam inter-related, and you may have already thought about this, should we open an issue to update smart_proxy_pulp to report back if crane is enabled?

manifests/crane/params.pp
stbenjam: It's the real default for the value, per https://docs.pulpproject.org/plugins/crane/index.html#configuration

If you're using it through pulp class, you probably want it to point to the pulp directory.

stbenjam:
> @stbenjam inter-related, and you may have already thought about this, should we open an issue to update smart_proxy_pulp to report back if crane is enabled?

Where would that be used?

ekohl:
@stbenjam Looks good. Squash and merge?

Issue #171, opened 2016-11-03
show_diff on /etc/pulp/server.conf show passwords


ccnifo:
During the first run or when applying a password change, _/etc/pulp/server.conf_ is modified and a diff is displayed showing the password, which may be logged. This may even lead the password to be exported and displayed in an external log analyzer. This applies to the default admin password, but more seriously to the MongoDB and message broker password.
The _/etc/pulp/server.conf_ resource should probably use the _show_diff = false_ parameter to avoid this potential leak.

stbenjam:
That seams reasonable, if you want to open a PR we'll get it merged quickly, otherwise it'll be when one of the devs gets around to doing it.

ccnifo:
I wrote some code providing the show_diff as a parameter. Given that's more like a security feature, I don't know if it's better to have it parametrized or to hardcode it?

stbenjam:
That looks ok to me, I would also add it to any of the importers in manifests/config.pp as they also contain passwords.

ccnifo:
4 files added, I'll open the PR, thanks.

PR #172, opened 2016-11-03
fixes GH-171 - add a show_conf_diff param, defaults to false

ccnifo:
These commits avoid displaying any diff by default on files containing passwords. They add a parameter to change this behaviour.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -9,11 +9,12 @@ class pulp::config {
   }
 
   file { '/etc/pulp/server.conf':
-    ensure  => file,
-    content => template('pulp/server.conf.erb'),
-    owner   => 'apache',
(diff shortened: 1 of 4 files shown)


ccnifo:
Given that there are 4 boolean params involved (enable_docker, enable_puppet, enable_rpm, show_conf_diff), thus 16 tests to write, that is a good motivation to put _show_diff = false_ as a default hardcoded value :) Especially as I'm not sure there are use cases for _show_conf_diff = true_
Do you agree?

stbenjam:
You don't need tests for every possible combination of parameters.

ccnifo:
OK, I have written just a few tests.

stbenjam:
@ccnifo Thanks! Looks good to me

PR #173, opened 2016-11-09
Refs #17298 - Add max tasks per Pulp worker

mbacovsky:
!!**Merge once Pulp >= 2.11 is added to Katello**!!

This feature requires Pulp >= 2.11. It allows setting of maximum number of tasks after which the celery worker gets recycled and releses its memory to the system.

Related patch to `puppet-katello` will follow

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -416,6 +416,10 @@ class pulp (
   validate_hash($additional_wsgi_scripts)
   validate_integer($max_keep_alive)
 
+  if $max_tasks_per_child {
+    validate_integer($max_tasks_per_child)
+  }
+
   if $https_cert {
     validate_absolute_path($https_cert)
   }
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -217,6 +217,10 @@
 # $enable_crane::               Boolean to enable crane docker repository
 #                               type:boolean
 #
+# $max_tasks_per_child::        Number of tasks after which the worker is restarted
+#                               and the memory it allocated is returned to the system
+#                               type:integer
+#
 # $enable_rpm::                 Boolean to enable rpm plugin. Defaults
 #                               to true
 #                               type:boolean
@@ -287,7 +291,7 @@
 #                               Warning: may display and log passwords contained in these files.
 #                               Defaults to false
 #                               type:boolean
-# 
+#
 class pulp (
   $version                   = $pulp::params::version,
   $crane_debug               = $pulp::params::crane_debug,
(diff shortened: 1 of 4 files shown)


manifests/init.pp
```diff
@@ -355,6 +359,7 @@
   $proxy_password            = $pulp::params::proxy_password,
   $num_workers               = $pulp::params::num_workers,
   $enable_crane              = $pulp::params::enable_crane,
+  $max_tasks_per_child       = $pulp::params::max_tasks_per_child,
```
ekohl: Would you mind calling ```validate_integer``` on this as well? Then we know it's an int and you can leave out the ```.to_i``` in the templates.

ekohl:
I assume this is still pending the new pulp version.

ehelms:
Yep.

SimonPe:
hey, I just double checked and this has no impact on pre 2.11 releases (it is just an environment var which doesn't get read)

It would be nice if this could get merged already for people who are already on 2.11 or 2.12

ehelms:
@SimonPe happy to merge if @mbacovsky handles the requests or if someone wnats to take over this PR

ehelms:
@mbacovsky We'd like to pull this in to the next release, do you have time to revisit?

ehelms:
Thanks for revisting this @mbacovsky looks like you have some test failures.

mbacovsky:
Sorry for not responding, I had muted notifications on this repo. 
@ekohl I've added the validation end removed conversion in the templates.
PR was rebased. Some of the tests still fail on places that seem unrelated to this PR.

mbacovsky:
Tests look good now. Though I'm not sure if the change in testing from `scope['pulp::max_tasks_per_child'] != :undef` to `!scope['pulp::max_tasks_per_child'].nil?` was the right fix. (I'm still a bit confused in puppet's :undefs and nils)

templates/systemd_pulp_workers
```diff
@@ -6,3 +6,11 @@ PULP_CONCURRENCY=<%= scope['pulp::num_workers'] %>
 
 # Configure Python's encoding for writing all logs, stdout and stderr
 PYTHONIOENCODING="UTF-8"
+
+# To avoid memory leaks, Pulp can terminate and replace a worker after processing X tasks. If
+# left commented, process recycling is disabled. PULP_MAX_TASKS_PER_CHILD must be > 0.
+<% if !scope['pulp::max_tasks_per_child'].nil? and scope['pulp::max_tasks_per_child'] > 0 %>
```
stbenjam: Generally we do this kind of check:

```erb
<% unless [nil, :undefined, :undef, ''].include?(scope.lookupvar('pulp::max_tasks_per_child')) -%>
```

Some examples:
https://github.com/Katello/puppet-pulp/blob/master/templates/server.conf.erb#L415

The reason it's like this is because of how different puppet versions handle undefined variables, it's a bit of a mess that'll be fixed once we can go to Puppet 4 only.

mbacovsky: @stbenjam thanks for explanantion. Updated.

mbacovsky:
Fixed the test in templates to work across puppet versions.

PR #180, opened 2016-12-10
version compared as number

PascalBourdier:


stbenjam:
Thanks @PascalBourdier!

PR #181, opened 2016-12-15
Fixes #16946 - add step to verify Pulp CA

chris1984:
Verified its working on downstream, since we do not have an EL6 build and that is the only way to reproduce on EL6

https://paste.fedoraproject.org/506976/48182110/

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -103,6 +103,11 @@ class pulp::config {
     }
   }
 
+    exec { 'run pulp-gen-ca':
+      command => '/usr/bin/pulp-gen-ca-certificate',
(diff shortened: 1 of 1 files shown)


manifests/config.pp
```diff
@@ -103,6 +103,11 @@
     }
   }
 
+    exec { 'run pulp-gen-ca':
+      command => '/usr/bin/pulp-gen-ca-certificate',
+      creates => '/etc/pki/pulp/ca.crt',
+}
```
stbenjam: The } should be aligned with the 'e' from exec

chris1984:
@stbenjam Fixed, thanks for the review :)

PR #182, opened 2016-12-19
Fixes #16941 - Check perms on Pulp key

chris1984:


manifests/config.pp
```diff
@@ -86,6 +86,13 @@
     mode    => '0644',
   }
 
+  file { '/etc/pki/pulp/rsa.key':
+    ensure => file,
+    owner  => 'root',
+    group  => 'apache',
+    mode   => '0640',
+        }
```
ekohl: Alignment is off.

chris1984:
@ekohl it looks to be an issue with an empty file, according to this:

https://github.com/pulp/pulp/blob/master/pulp.spec#L906-L919

What are your suggestions?

stbenjam:
This looks like where it's actually created:
https://github.com/pulp/pulp/blob/master/pulp.spec#L476-L493

I think that because pulp::config should run after pulp::install, the file will always exist by the time puppet runs here, so puppet will only fix the permissions (what we want).

stbenjam:
To be honest though, I'm a little skeptical of this change, there are tens of thousands of files managed by packages that Foreman, Katello, and others use - puppet can't maintain permissions on them all.  

This was a one-off case of a user flailing their arms trying to fix something and making the situation worse, should we handle all of those cases? I want to say no, but also these keys do seem to be something different since they're generated in %post scriplets in the RPM and `rpm -V` doesn't seem to detect changes.  I think maybe something to do with the `%ghost` directive:

```
[root@sat-rhel7 ~]# chown root:root /etc/pki/pulp/rsa.key 
[root@sat-rhel7 ~]# rpm -V pulp-server | grep rsa.key
[root@sat-rhel7 ~]# 
```

ekohl:
You can replicate the post in puppet:

```puppet
    exec { 'create pulp private key':
      command => "openssl genrsa -out ${::pulp::rsa_key} 2048",
      creates => $::pulp::rsa_key,
      umask   => '770',
    } ->
    exec { 'create pulp public key':
      command => "openssl rsa -in ${::pulp::rsa_key} -pubout > ${::pulp::rsa_pub}",
      creates => $::pulp::rsa_pub,
      umask   => '770',
    }
  
    file { $::pulp::rsa_key:
      ensure   => file,
      owner    => 'root',
      group    => 'apache',
      mode     => '0640',
      requires => Exec['create pulp private key'],
    }
```

stbenjam:
I wouldn't replicate the the %post scriplet, what if it changes in the future?

ekohl:
Maybe an even simpler solution: if you remove the ```ensure => file``` puppet will only ensure the permissions and not create it.

manifests/config.pp
```diff
@@ -86,6 +86,13 @@
     mode    => '0644',
   }
 
+  file { '/etc/pki/pulp/rsa.key':
```
ekohl: We already have this path as ```$::pulp::rsa_key``` so that would be preferred.

manifests/config.pp
```diff
@@ -86,6 +86,13 @@
...
+    ensure => file,
```
ekohl: If you remove this, then puppet won't create the file and thus not interfere with the ```%post```.

stbenjam:
> Maybe an even simpler solution: if you remove the ensure => file puppet will only ensure the permissions and not create it.

That seems better, but I'm not sure I want to get in the business of managing permissions on packaged files.

ekohl:
IMHO this is a bit of a grey area. It is in post so it's not really managed, just a best effort. It's also in /etc so I could accept this (given the changes I suggested) but I don't know how common this is. I'll leave it up to you @stbenjam.

stbenjam:
It is tracked in the RPM spec files, with a permission on it: 
  https://github.com/pulp/pulp/blob/master/pulp.spec#L442 (see like 439 above)

I feel like it's more a bug in RPM that it doesn't track this in verify (-V).

@chris1984 But, it's fine - if you want to remove the `ensure => file` as @ekohl suggested we can take this.

stbenjam:
Thanks @chris1984

PR #186, opened 2017-01-05
add the posibility to install katello_agent

cristifalcas:
The package is in katello repo, so the user will have to make it available somehow

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -17,6 +17,10 @@ class pulp::install {
     )
   }
 
+  if $pulp::enable_katello {
+    package { ['pulp-katello']: ensure => $pulp::version, }
+  }
+
   if $pulp::enable_parent_node {
     package { ['pulp-nodes-parent']: ensure => $pulp::version, }
   }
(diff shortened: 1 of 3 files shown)


ehelms:
This looks like it doesn't install katello-agent but rather our pulp-katello plugin which is really just a yum_clone distributor (with a poorly named repository).

cristifalcas:
Ah, yes. It's for installing pulp-katello package.

manifests/install.pp
```diff
@@ -17,6 +17,14 @@
     )
   }
 
+  if $pulp::enable_katello {
+    if $pulp::messaging_transport == 'rabbitmq' {
+      fail('pulp-katello plugin doesn\'t support rabbitmq')
```
ehelms: I don't think this is true? Given that the `pulp-katello` plugin is just adding a distributor type to Pulp.

cristifalcas: Correct. I thought initially that it is connected to katello somehow via qpid.

ehelms:
Looking good to me - mind squashing the commits and I'll merge?

ehelms:
Thanks @cristifalcas !

ehelms:
@cristifalcas If you need an official release with your new provides and updates just let me know

PR #187, opened 2017-01-05
add an schedule provider

cristifalcas:
Copied from previous PR:

This is an abomination because of some limiting factors: 
- I can't make the connection between repo id + schedule to a resource name
- I don't think that multiple schedules per repo can be implemented (from my point of view it can't be done)

Limitations:
- we can't have multiple schedules for the same repo, so we force the existence of only one schedule per repo (the one with the smallest sync_id). The resource name will be the repo_id
- user should not set the name of a repo to be the same as an existing schedule_id (probability is very small)
- if the repo (resource name) doesn't exists, the resource will fail (normal)
- there is a global variable used ($schedules_info)

Explanation on how it's implemented:
- when we get the resources, we set the resource name to repo_id only for the first schedule
- the other schedules get the name = schedule_id (this is useful only for purging)
- we need to keep somewhere the schedule_id for the first ones ($schedules_info['schedule_id']): this is needed in update
- we need to keep the repo_id for the second ones ($schedules_info['repo_id']): needed in delete, when we purge the resources
- since this provider is global for all repos, we need to keep the repo_type for all repos ($schedules_info['repo_type'])
- a special case is when we have a single resource defined in puppet with ensure => absent: we keep the schedule id in $schedules_info['repo_id']

Example usage:

```
pulp_schedule { 'scl_ruby22_el7':
  ensure        => 'present',
  enabled       => 'true',
  schedule_time => '2000-W01-6T12:00Z/P1W',
}

  pulp_schedule { 'katello_client_el7':
    ensure        => absent
  }
```

There is still a problem when purging repos and schedules: if the repo is deleted first, the schedule will fail to do the purging. I think this is because when deleting the repo, pulp cleans the schedules also.

I don't know how to fix this yet. I could to an extra check in delete to see if the schedule still exists, but it looks ugly. Also I don't like to add more special cases then there are already.

### DIFF
--- a/lib/puppet/provider/pulp_schedule/api.rb
+++ b/lib/puppet/provider/pulp_schedule/api.rb
@@ -1,0 +1,167 @@
+#  Limitations:
+# - we can't have multiple schedules for the same repo, so we force the existence of only one schedule per repo 
+#   (the one with the smallest sync_id). The resource name will be the repo_id
+# - user should not set the name of a repo to be the same as an existing schedule_id (probability is very small)
+# - if the repo (resource name) doesn't exists, the resource will fail (normal)
+# - there is a global variable used ($schedules_info)
+#
+# Explanation on how it's implemented:
+# - when we get the resources, we set the resource name to repo_id only for the first schedule
+# - the other schedules get the name = schedule_id (this is useful only for purging)
+# - we need to keep somewhere the schedule_id for the first ones ($schedules_info['schedule_id']): this is needed in update
+# - we need to keep the repo_id for the second ones ($schedules_info['repo_id']): needed in delete, when we purge the resources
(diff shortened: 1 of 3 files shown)


ehelms:
Have you chatted with the Pulp team about these limitations? Would it be worth it to get @mhrivnak from the Pulp team to take a look and offer some guidance?

cristifalcas:
This is not a pulp issue, but maybe a limitation regarding my knowledge about puppet. Currently the namevar is repo-id, so this is why with this provider we can only have one schedule per repo. 

I don't know how o implement a provider with multiple namevars. If you can point me to an other provider that implements multiple namevars, to see how it's done, maybe I will manage to do it here also.

ehelms:
Oh I see now. I do not know of one. I'll merge this as is for now and if we come across it can always update.

Thanks @cristifalcas !

PR #189, opened 2017-01-09
Fixes #16253 - Add max speed var to Katello

chris1984:


chris1984:
2nd part of PR for Katello https://github.com/Katello/puppet-katello/pull/162

manifests/init.pp
```diff
@@ -203,6 +203,8 @@
 #
 # $proxy_password::             Proxy password for authentication
 #
+# $max_speed::                  Maximum bandwidth used per download in bytes/sec from CDN
```
stbenjam: Is this setting only relevant for yum importer? I think the top level variable name should indicate it, maybe `$yum_max_seed`.  

This also applies to all yum repos synced by pulp, right? So I'd avoid saying CDN, which in the context of pulp probably only applies to Red Hat.

templates/yum_importer.json.erb
```diff
@@ -15,8 +15,13 @@
...
-    "proxy_password": "<%= scope['pulp::proxy_password'] %>"
+    "proxy_password": "<%= scope['pulp::proxy_password'] %>",
     <% else %>
-    "proxy_password": null
+    "proxy_password": null,
+    <% end %>
+    <% unless [nil, :undefined, :undef, ''].include?(scope['pulp::max_speed']) -%>
+    "max_speed": <%= scope['pulp::max_speed'] %>"
+    <% else %>
+    "max_speed": null
```
stbenjam: Is this needed, or could it just be omitted entirely?

ekohl: Given JSON doesn't support trailing commas I think this may be easier.

chris1984:
@stbenjam when you get a chance can I get a review on this. It cant be merged until this PR gets in as well though https://github.com/Katello/puppet-katello/pull/162 but at least can focus my efforts on the other knowing this one is good.

manifests/init.pp
```diff
@@ -203,6 +203,8 @@
 #
 # $proxy_password::             Proxy password for authentication
 #
+# $yum_max_speed::              Maximum bandwidth used per download in bytes/sec
```
stbenjam: I would add `type:integer`, and include @ekohl's suggestion validation, but otherwise this PR looks good to me

chris1984:
@stbenjam @ekohl Thanks again for the review and help with this. Can you look it over again one last time please? It cant be merged until this https://github.com/Katello/puppet-katello/pull/162 is fixed but at least 1 part of this is taken care of.

sean797:
I would love if this could use https://github.com/puppetlabs/puppetlabs-stdlib#to_bytes. It would be a lot easier for users.

manifests/init.pp
jlsherrill: this is also per worker right?  we may want to make that clear in the description

sean797: @jlsherrill it's per task :-)

jlsherrill: right, i guess i'm thinking that most people use this to limit their bandwidth and might not know what 'per task' means, granted they likely wouldn't know what 'per worker' means either, but the maximum it would take up is the speed times the number of workers (which would be the number of tasks executing at once)

manifests/init.pp
```diff
@@ -420,6 +424,11 @@
     validate_string($ssl_protocol)
   }
 
+  if $yum_max_speed {
+    validate_integer($yum_max_speed)
+    to_bytes($yum_max_speed)
```
chris1984: @sean797 I am not sure if this is where its supposed to go, can you check?

manifests/init.pp
```diff
@@ -203,6 +203,9 @@
 #
 # $proxy_password::             Proxy password for authentication
 #
+# $yum_max_speed::              Maximum bandwidth used per download in kb/sec per worker
```
chris1984: @sean797 @jlsherrill I changed this to KB since we are using the to_bytes that Sean suggested. Will that work for the description?

manifests/init.pp
sean797: That line looks okay, but we would expect users to supply something like "4 Kb", "60 Mb" or "1 Gb" so there is no need to `validate_integer` in the line above, could use `validate_string` if you would like.

manifests/init.pp
```diff
@@ -203,6 +203,9 @@
...
+#                               type:integer
```
sean797: Ditto

manifests/init.pp
sean797: Have a look at https://docs.pulpproject.org/plugins/pulp_rpm/tech-reference/yum-plugins.html?highlight=max_speed#configuration-parameters it says "The maximum download speed in bytes/sec for a task (such as a sync); defaults to None"

As were allowing users to put something like "4 Kb" in maybe something like "The maximum download speed for a Pulp task, such as a sync. (e.g. "4 Kb" or 1Gb")" would be better?

chris1984:
@sean797 can i get a rereview please. Once this one looks good ill update the fixtures on the other pr and make changes and see if tests pass.

manifests/init.pp
```diff
@@ -420,6 +424,11 @@
     validate_string($ssl_protocol)
   }
 
+  if $yum_max_speed {
+    validate_string($yum_max_speed)
+    to_bytes($yum_max_speed)
```
stbenjam: I don't think this works, $yum_max_speed will still contain a string.  You should do something like

```
$real_yum_max_speed = to_bytes($yum_max_speed)
```

And then use $real_yum_max_speed in the template

chris1984:
@stbenjam Can you help me find what its upset about, also does the template look ok?

manifests/init.pp
```diff
@@ -420,6 +424,12 @@
     validate_string($ssl_protocol)
   }
 
+  if $yum_max_speed {
+    validate_string($yum_max_speed)
+    $real_yum_max_speed = to_bytes($yum_max_speed)
+  else { $real_yum_max_speed = undef }
+}
```
stbenjam: This syntax isn't valid.

```puppet
if $yum_max_speed {
  validate_string($yum_max_speed)
  $real_yum_max_speed = to_bytes($yum_max_speed)
}  else {
  $real_yum_max_speed = undef
}
```

templates/yum_importer.json.erb
```diff
@@ -15,8 +15,13 @@
     "proxy_username": null,
     <% end %>
     <% unless [nil, :undefined, :undef, ''].include?(scope['pulp::proxy_password']) -%>
-    "proxy_password": "<%= scope['pulp::proxy_password'] %>"
+    "proxy_password": "<%= scope['pulp::proxy_password'] %>",
     <% else %>
-    "proxy_password": null
+    "proxy_password": null,
+    <% end %>
+    <% unless [nil, :undefined, :undef, ''] $real_yum_max_speed -%>
```
stbenjam: Ditto, this isn't valid syntax, you need to do the scope lookup like was here before

templates/yum_importer.json.erb
```diff
@@ -15,8 +15,13 @@
...
+    "max_speed": <%= $real_yum_max_speed %>"
```
stbenjam: Same here

chris1984:
@sean797 @stbenjam Made the corrections requested. Can I get a rereview please... Thanks.

templates/yum_importer.json.erb
```diff
@@ -15,8 +15,13 @@
     "proxy_username": null,
     <% end %>
     <% unless [nil, :undefined, :undef, ''].include?(scope['pulp::proxy_password']) -%>
-    "proxy_password": "<%= scope['pulp::proxy_password'] %>"
+    "proxy_password": "<%= scope['pulp::proxy_password'] %>",
     <% else %>
-    "proxy_password": null
+    "proxy_password": null,
+    <% end %>
+    <% unless [nil, :undefined, :undef, ''].include?(scope[' $real_yum_max_speed']) -%>
```
stbenjam: Use `scope['pulp::real_yum_max_speed'] `, it follows the pattern of all the other uses of `scope` in this file.    i.e., no $, and definitely no spaces.

The documentation is here https://docs.puppet.com/puppet/latest/lang_template_erb.html#scopevariable-or-scopelookupvarvariable

chris1984:
Stephen thanks for the link https://docs.puppet.com/puppet/latest/lang_template_erb.html#scopevariable-or-scopelookupvarvariable me-- should have read the docs. 

I can see now that it can see all the vars even ones I did not define in params.pp So no need to use the $ since it can find it based on pulp::var, makes more sense now

@stbenjam @sean797 hopefully this one is a good to go.

PR #190, opened 2017-01-11
python-gofer-qpid and python-gofer-amqp version

PascalBourdier:
python-gofer-qpid and python-gofer-amqp are not yet available in the public repo in the same version than pulp
They seem to be in different versions


cristifalcas:
We rely on the current implementation to keep those packages updated. Can there still be a parameter for this? Something like $pulp::other_version that can be set to undef/latest/installed only?

ehelms:
Agreed, if they need to be configured individually, parameters for each (and put into the advanced parameters documentation) seems best.

ekohl:
https://github.com/Katello/puppet-pulp/pull/195 has been merged and allows you to specify the versions.

PR #191, opened 2017-01-18
fixes #18144 - correct set crane data_dir

thomasmckay:


manifests/crane/params.pp
```diff
@@ -6,6 +6,6 @@
   $debug = false
   $port = 5000
 
-  $data_dir = '/var/lib/crane/metadata'
+  $data_dir = '/var/lib/pulp/published/docker/v2/app'
```
jlsherrill: so i think this was set that way because that is the default in crane: https://github.com/pulp/crane/blob/master/crane/data/default_config.conf

we likely need to override this parameter.

jlsherrill: I think you'd add a line here: https://github.com/Katello/puppet-katello/blob/master/manifests/init.pp#L124

$crane_data_dir => '/var/lib/pulp/published/docker/v2/app'

parthaa: Well we had this value before puppet-crane got deprecated. Look at => https://github.com/Katello/puppet-crane/blob/master/manifests/params.pp#L8 

Dont see why we need to override ?

PR #193, opened 2017-02-13
refs #17367 - allow pulp to listen on multiple ports

stbenjam:


ehelms:
Does the puppetlabs-apache version range we have this pinned to support arrays of ports?

stbenjam:
No released version of puppetlabs-apache has it, and they're slow to release.  I'll have to create a 1.11.x branch on the Foreman fork https://github.com/theforeman/puppetlabs-apache and cherry-pick it, and then hopefully ask Foreman nicely to pin it.

manifests/init.pp
```diff
@@ -102,11 +102,13 @@
...
-# $https_cert::                 apache public certificate for ssl
+# $https_cert::                 Apache public certificate for ssl
 #
-# $https_key::                  apache private certificate for ssl
+# $https_key::                  Apache private certificate for ssl
 #
-# $https_chain::                apache chain file for ssl
+# $https_chain::                Apache chain file for ssl
+#
+# $https_ports::                Ports to listen on
```
ekohl: Not sure if it's used, but maybe add ```type:array```?

ehelms:
@stbenjam any updates on puppetlabs-apache release or forked cherry pick?

ekohl:
Now that puppetlabs-apache 2.0.0 is out we can rely on this functionality.

ekohl:
I don't see how this will solve the original (related issue). https://github.com/Katello/puppet-foreman_proxy_content/blob/3a1118deb9104e167c754aa57914eb8894cb54e9/manifests/init.pp#L211-L213 does a similar thing, but only proxies certain URLs. Can't we expand that to include the RHSM URLs? See https://github.com/Katello/puppet-foreman_proxy_content/blob/3a1118deb9104e167c754aa57914eb8894cb54e9/templates/_pulp_gpg_proxy.erb as well.

Closing this now, but can be re-opened later if I misunderstood this.

stbenjam:
This wasn't a standalone fix, it was to go with https://github.com/Katello/puppet-foreman_proxy_content/pull/113.

We need to move RHSM to 443, but give users the option to keep 8443 for backwards compatability - since there's Katello's and Satellite's in the wild with tends of thousands of registered hosts.

ekohl:
That makes sense. Thanks for documenting that, wasn't obvious from just the code.

PR #195, opened 2017-02-14
Messaging transport version

khdevel:
[messaging_transport_version]
    - new parameter 'messaging_transport_version'
    - it is related to the issue that when $version is set to e.g. 2.12.0-1.el7 it is also assigned
    to packages like 'python-gofer-qpid', 'python-gofer-amqp' or 'gofer' which do not have such number
    
     Changes to be committed:
    - modified:   manifests/init.pp
    - modified:   manifests/params.pp
    - modified:   manifests/install.pp
    - modified:   manifests/consumer.pp
    - modified:   manifests/consumer/install.pp
    - modified:   manifests/consumer/params.pp

### DIFF
--- a/manifests/consumer.pp
+++ b/manifests/consumer.pp
@@ -88,6 +88,8 @@
 #
 # $messaging_transport::           The AMQP transport name. Valid options are 'qpid' or 'rabbitmq'. The default is 'qpid'.
 #
+# $messaging_version::             Determines the version of packages related to the 'messaging transport protocol'.
+#
 # $messaging_cacert::              The (optional) absolute path to a PEM encoded CA certificate to validate the identity of the
 #                                  broker.
 #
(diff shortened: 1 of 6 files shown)


manifests/consumer.pp
```diff
@@ -88,6 +88,8 @@
 #
 # $messaging_transport::           The AMQP transport name. Valid options are 'qpid' or 'rabbitmq'. The default is 'qpid'.
 #
+# $messaging_version::             Determines the version of packages related to the 'messaging transport protocol'.
```
ekohl: You're documenting a parameter that doesn't exist in this class.

khdevel: I have updated the code...

ehelms:
@ekohl They come from the gofer package so they should be in sync -- here is an example http://koji.katello.org/koji/buildinfo?buildID=30583

ekohl:
@ehelms do you then agree this can be merged?

ekohl:
Thanks!

khdevel:
Thank you very much!

Issue #200, opened 2017-03-07
selinux prevents httpd binding to 5000


rgarrigue:
Hi

The module couldn't start httpd

```
==> test.vagrant.local: Error: Could not start Service[httpd]: Execution of '/bin/systemctl start httpd' returned 1: Job for httpd.service failed because the control process exited with error code. See "systemctl status httpd.service" and "journalctl -xe" for details.
==> test.vagrant.local: Error: /Stage[main]/Apache::Service/Service[httpd]/ensure: change from stopped to running failed: Could not start Service[httpd]: Execution of '/bin/systemctl start httpd' returned 1: Job for httpd.service failed because the control process exited with error code. See "systemctl status httpd.service" and "journalctl -xe" for details.
```

Because httpd couldn't use port 5000

```
[root@test vagrant]# journalctl -u httpd
-- Logs begin at Tue 2017-03-07 14:59:06 CET, end at Tue 2017-03-07 15:14:34 CET. --
Mar 07 15:09:24 test.vagrant.local systemd[1]: Starting The Apache HTTP Server...
Mar 07 15:09:24 test.vagrant.local httpd[29148]: (13)Permission denied: AH00072: make_sock: could not bind to address [::]:5000
Mar 07 15:09:24 test.vagrant.local httpd[29148]: (13)Permission denied: AH00072: make_sock: could not bind to address 0.0.0.0:5000
Mar 07 15:09:24 test.vagrant.local httpd[29148]: no listening sockets available, shutting down
Mar 07 15:09:24 test.vagrant.local httpd[29148]: AH00015: Unable to open logs
Mar 07 15:09:24 test.vagrant.local systemd[1]: httpd.service: main process exited, code=exited, status=1/FAILURE
Mar 07 15:09:24 test.vagrant.local kill[29150]: kill: cannot find process ""
Mar 07 15:09:24 test.vagrant.local systemd[1]: httpd.service: control process exited, code=exited status=1
Mar 07 15:09:24 test.vagrant.local systemd[1]: Failed to start The Apache HTTP Server.
Mar 07 15:09:24 test.vagrant.local systemd[1]: Unit httpd.service entered failed state.
Mar 07 15:09:24 test.vagrant.local systemd[1]: httpd.service failed.
```

... because selinux is a wonderful security feature

```
[root@test vagrant]# grep httpd /var/log/audit/audit.log | grep denied
[...]

PR #204, opened 2017-03-08
Allow apache to bind to port 5000 under SELinux

rgarrigue:
Allow apache httpd binding to 5000/tcp

Solve #200

### DIFF
--- a/spec/classes/pulp_apache_spec.rb
+++ b/spec/classes/pulp_apache_spec.rb
@@ -41,6 +41,8 @@ describe 'pulp::apache' do
         :wsgi_import_script      => '/usr/share/pulp/wsgi/webservices.wsgi',
       })
     end
+    
+    it { is_expected.to contain_selinux__boolean('httpd_use_openstack') }
   end
 
   context 'with parameters' do


manifests/apache.pp
```diff
@@ -123,4 +123,6 @@
     group   => 'root',
     mode    => '0644',
   }
+  
+  selinux::boolean { 'httpd_use_openstack': }
```
ekohl: We're not using a module for selinux. That means we could use ```selboolean``` which is built in, but then you need to be sure that selinux is enabled (either enforcing or permissive).

It's also only needed for crane so it belongs in ```pulp::crane::apache```.

ekohl:
@rgarrigue could you have a look at my comment?

rgarrigue:
@ekohl saw that. Will see if I can fix this PR this week.

ekohl:
Btw, I think the reason we're not hitting this in the Katello scenario is that katello-selinux takes care of this.

sean797:
Should this be done in https://github.com/pulp/pulp/blob/master/server/selinux/server/pulp-server.te instead of katello-selinux?

ekohl:
I do think that would make sense but not an expert on selinux. Going to close this since @rgarrigue removed his fork so I think it's unlikely he's going to work on this.

PR #208, opened 2017-03-16
Fixes #17219 - bump squid3 to 1.0.2

Klaas-:
@ehelms in f461a2113dcd1aaa59fd2cd5d7cd461b20bad2f3 you pinned it to 1.0.0 - but I don't really see why the updates should break el6 - in any case if the problem persists put this off till el6 support is dropped from katello :D

Greetings
Klaas

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -33,7 +33,7 @@
     },
     {
       "name": "thias/squid3",
-      "version_requirement": "1.0.0"
+      "version_requirement": "1.0.2"
     }
   ],
   "requirements": [


ehelms:
Take a look at https://github.com/thias/puppet-squid3/pull/50 (which is still unmerged :/). Some of the Katello modules have dropped EL6 support, this one appears to have as well looking at the metadata.json so we should be fine. I will proceed once you get a chance to look at that PR.

Klaas-:
@ehelms I don't have a el6 to test if that persists, it doesn't seem to be fixed and the guy maintaining the module is quite bad with PRs ... :) in any case on el7 that dir is part of the squid package so I see no problem with merging this if the module has already dropped el6 support

ehelms:
Agreed - thanks @timogoebel !

PR #209, opened 2017-03-16
broker service may not run on the same node

timogoebel:
When installing pulp and the message broker on different nodes, the
massage broker service is not present in the catalog and catalog
compilation fails.
Using a resource collector instead of referencing the service directly
makes the `Service[$broker_service]` optional.

### DIFF
--- a/manifests/broker.pp
+++ b/manifests/broker.pp
@@ -20,8 +20,8 @@ class pulp::broker {
     }
   }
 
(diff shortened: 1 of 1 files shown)


ehelms:
Looks good once tests are passing

timogoebel:
@ehelms : Tests passed, now.

ehelms:
Thanks @timogoebel !

PR #211, opened 2017-03-21
allow to enable/disable redirects.

cristifalcas:

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -176,6 +176,9 @@
 #
 # $manage_squid::               boolean controls whether squid configuration is managed or manual
 #
+# $lazy_enabled::               This controls whether redirect is enabled
+#                               type:boolean
+#
 # $lazy_redirect_host::         The host FQDN or IP to which requests are redirected.
 #
 # $lazy_redirect_port::         The TCP port to which requests are redirected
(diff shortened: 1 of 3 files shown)


templates/server.conf.erb
```diff
@@ -412,6 +412,9 @@ enabled: <%= scope['pulp::email_enabled'] %>
 #   downloading content from the Squid cache.
 
 [lazy]
+<% unless [nil, :undefined, :undef, ''].include?(scope.lookupvar('pulp::lazy_enabled')) -%>
```
ekohl: Since this is validated as a bool this check isn't needed.

manifests/init.pp
```diff
@@ -176,6 +176,9 @@
 #
 # $manage_squid::               boolean controls whether squid configuration is managed or manual
 #
+# $lazy_enable::                This controls whether redirect is enabled
```
ekohl: To me it's a bit confusing that first it's lazy, then it's redirect. You also have ```$lazy_enable``` where the code has ```$lazy_enabled``` (note the missing d)

templates/server.conf.erb
cristifalcas: This is more for undefined values.
Should I remove it?

manifests/init.pp
cristifalcas: The explanation is from server.conf:

# enabled:
#   This controls whether redirect is enabled

templates/server.conf.erb
ekohl: Yes. We know it's never any of those values.

ehelms:
I asked the Pulp team about this config option and they indicated they were not aware of it. Can you point out where it comes from?

ehelms:
I should also say I checked https://github.com/pulp/pulp/blob/master/server/etc/pulp/server.conf#L391 first and when I didn't see it in their official server.conf I reached out to them.

cristifalcas:
It was based on the parameter in the server.conf template from the puppet module. If it doesn't exists, can the lazy streaming be disabled somehow?

cristifalcas:
ok, it seems it's a pulp repo parameter.

PR #212, opened 2017-03-22
check broker service only when we manage it

cristifalcas:


ekohl:
The way I read this was that it automatically does the chaining only if the service is defined. @timogoebel  is that correct?

timogoebel:
Yes, I don't see the value of this change. @cristifalcas : Would you mind elaborating a bit why you propose this change?

cristifalcas:
You are right. I have an internal module that was failing and was complaining because it doesn't know the service. Or something like this, because I can't reproduce it anymore. Most probably was from something else.

Sorry for the noise.

PR #216, opened 2017-03-31
Fix puppet-lint arrow_on_right_operand_line

alexjfisher:
This check was added in puppet-lint 2.2.0.
https://github.com/rodjek/puppet-lint/pull/672
https://docs.puppet.com/puppet/latest/style_guide.html#chaining-arrow-syntax

### DIFF
--- a/manifests/service.pp
+++ b/manifests/service.pp
@@ -3,8 +3,8 @@ class pulp::service {
   exec { 'pulp refresh system service':
     command     => '/bin/systemctl daemon-reload',
     refreshonly => true,
-  } ->
-  service { ['pulp_celerybeat', 'pulp_workers', 'pulp_resource_manager', 'pulp_streamer']:
+  }
+  -> service { ['pulp_celerybeat', 'pulp_workers', 'pulp_resource_manager', 'pulp_streamer']:
     ensure => running,
     enable => true,
   }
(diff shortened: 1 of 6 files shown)


ekohl:
We've been talking about the arrow syntax and we're leaning to ignoring it. I don't think any of us like it and syntax highlighters also have issues with it (see here on github and vim at least).

alexjfisher:
@ekohl Ok.  I'm mostly indifferent, but I guess puppet-lint should follow the official styleguide for its defaults.  To disable the warning, are you going to update all Foreman/Katello modules via a modulesync?

manifests/service.pp
```diff
@@ -3,8 +3,8 @@
   exec { 'pulp refresh system service':
     command     => '/bin/systemctl daemon-reload',
     refreshonly => true,
-  } ->
-  service { ['pulp_celerybeat', 'pulp_workers', 'pulp_resource_manager', 'pulp_streamer']:
+  }
+  -> service { ['pulp_celerybeat', 'pulp_workers', 'pulp_resource_manager', 'pulp_streamer']:
```
alexjfisher: @ekohl I agree this looks particularly bad.  All the other changes look fine either way to me.

ekohl:
@alexjfisher we have to agree on something, then we either do a a modulesync to ignore it or bulk fix. theforeman/* were already updated in bulk.

ekohl:
The discussion was happening in parallel on #theforeman-dev. We agreed on https://github.com/Katello/foreman-installer-modulesync/pull/23 for now, at least until we get some larger refactoring PRs in that are already in progress and would get bad merge conflicts.

PR #219, opened 2017-04-03
Fix repo_auth parameter

alexjfisher:
repo_auth was not in scope in config.pp so the setting was always false
in the template.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -127,8 +127,7 @@
 #                               should be changed once the server is operational
 #
 # $repo_auth::                  Boolean to determine whether repos managed by
-#                               pulp will require authentication. Defaults
-#                               to true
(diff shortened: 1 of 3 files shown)


manifests/init.pp
```diff
@@ -128,7 +128,7 @@
 #
 # $repo_auth::                  Boolean to determine whether repos managed by
 #                               pulp will require authentication. Defaults
-#                               to true
+#                               to false
```
ekohl: We've started removing the defaults from documentation because they tend to do out of date (like you see here). Would you mind removing it?

alexjfisher: removed

stbenjam:
Thanks! Sorry for the delay on getting your PR's in @alexjfisher

PR #221, opened 2017-04-03
LDAP authentication support

alexjfisher:
For those who want to implement
https://docs.pulpproject.org/user-guide/authentication.html#ldap-whole-api-example

### DIFF
--- a/spec/classes/pulp_apache_spec.rb
+++ b/spec/classes/pulp_apache_spec.rb
@@ -393,6 +393,36 @@ Alias /pulp/nodes/content /var/www/pulp/nodes/content
 ')
       end
     end
+
+    describe 'with ldap parameters' do
+      let :pre_condition do
+        "class {'pulp':
+           ldap_url           => 'ldaps://ad.example.com?sAMAccountName',
+           ldap_bind_dn       => 'cn=pulp,dc=example,dc=com',
+           ldap_bind_password => 'BIND_PASSWORD',
+          }"
+      end
+
+      it 'should configure apache for LDAP authentication' do
+        verify_concat_fragment_contents(catalogue, 'pulp-https-directories', [
+          '  <Files "webservices.wsgi">',
+          '    SetEnvIfNoCase ^Authorization$ "Basic.*" USE_APACHE_AUTH=1',
+          '    Order allow,deny',
+          '    Allow from env=!USE_APACHE_AUTH',
+          '    Satisfy Any',
+          '    AuthType basic',
+          '    AuthBasicProvider ldap',
+          '    AuthName "Pulp"',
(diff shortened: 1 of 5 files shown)


spec/classes/pulp_apache_spec.rb
```diff
@@ -393,6 +393,26 @@
...
+      let :pre_condition do
+        "class {'pulp':
+           ldap_url           => 'ldaps://ad.example.com?sAMAccountName',
+           ldap_bind_dn       => 'cn=pulp,dc=example,dc=com',
+           ldap_bind_password => 'BIND_PASSWORD',
+          }"
+      end
+
+      it 'should configure apache for LDAP authentication' do
+        is_expected.to contain_concat__fragment('pulp-https-directories')
```
ekohl: We have a ```verify_concat_fragment_exact_contents``` helper for this.

alexjfisher: I've used the other helper `verify_concat_fragment_contents`.  Hope that's ok.

templates/ldap_custom_fragment.erb
```diff
@@ -0,0 +1,15 @@
...
+    SetEnvIfNoCase ^Authorization$ "Basic.*" USE_APACHE_AUTH=1
+    Order allow,deny
+    Allow from env=!USE_APACHE_AUTH
+    Satisfy Any
+
+    # configure basic auth
+    AuthType basic
+    AuthBasicProvider ldap
+    AuthName "Pulp"
+    AuthLDAPURL "<%= scope['pulp::ldap_url'] %>"
```
ekohl: Isn't this possible to pass in as ```auth_ldap_url``` to the directories options?

alexjfisher: It was possible to configure some lines with directories options, but not others.  custom_fragments are added at the end of the template.  Mixing custom_fragment with other options ended up quite messy.  At least this way, the output matches the [documented ](https://docs.pulpproject.org/user-guide/authentication.html#ldap-whole-api-example)way of configuring this feature.

ekohl: That makes sense and I suspected it was something like this.

stbenjam:
It looks OK to me, I don't have an easy way to test the actual functionality at the moment though.

alexjfisher:
Not sure what to suggest.  Are you happy that the PR doesn't change module behaviour unless an `ldap_url` is provided?  I've been using this PR with active directory for the last few weeks.

ekohl:
I'm happy to accept it. The tests show it's the same as the documentation, doesn't change existing behaviour and I trust @alexjfisher's use in production.

ekohl:
Thanks!

PR #224, opened 2017-04-12
Use concat 2.2.1 in .fixtures

alexjfisher:
concat master has just dropped support for puppet 3.  2.2.1 was made a
few days earlier.  Presumably this will be the last release to support
puppet 3 and the next release of concat will be 3.0.0.


ekohl:
Thanks

PR #227, opened 2017-04-12
Refactor pulp::child and use puppetlabs-apache vhost options

ekohl:
This needs testing. I'm mostly wondering what you think of the pattern.

I also noticed that pulp::child has unused parameters which makes me think this is not usuable as a stand alone class while the rest implies it does.


manifests/child/apache.pp
```diff
@@ -0,0 +1,39 @@
...
+    directories            => $directories,
+    ssl                    => true,
+    ssl_cert               => $ssl_cert,
+    ssl_key                => $ssl_key,
+    ssl_ca                 => $ssl_ca,
+    ssl_verify_client      => 'optional',
+    ssl_options            => '+StdEnvVars',
+    ssl_verify_depth       => '3',
+    # allow older yum clients to connect, see bz 647828
+    custom_fragment        => 'SSLInsecureRenegotiation On',
```
stbenjam: I would guess EL5.  There's still a lot of CentOS 5 boxes out there - how about making it parameterized and default off?

ekohl: That sounds fine to me.

ekohl:
I'll revisit this when https://github.com/Katello/puppet-katello/pull/189 has been merged.

manifests/child.pp
```diff
@@ -4,22 +4,18 @@
...
-  $parent_qpid_scheme    = 'ssl',
-  $parent_qpid_port      = '5671',
-  $oauth_effective_user  = 'admin',
-  $oauth_key             = 'key',
-  $oauth_secret          = 'secret',
-  $ssl_cert              = '/etc/pki/pulp/ssl_apache.crt',
-  $ssl_key               = '/etc/pki/pulp/ssl_apache.key',
-  $server_ca_cert        = '/etc/pki/pulp/ca.crt'
+  String[1] $parent_fqdn,
+  String $oauth_effective_user = 'admin',
```
ekohl: Note I left these in, but they are *not* used. ```pulp::child::config``` uses values from the main pulp class. I'm wondering if that's a bug

ekohl:
I'm wondering where this code is even used. I can't really find where this is included in the installer.

stbenjam:
We don't use `pulp::child` in Katello, all the pulps are basically masters since 3.1 I think.  It's still a feature of pulp that people may want to use though.

manifests/child.pp
stbenjam: It may be, I think pulp::child is intended to be used standalone so it should use it's own values.

ekohl:
I'm leaning to merging this and changing things in incompatible ways in another PR.

ekohl:
Updated: added a trailing comma in child.pp.

Issue #230, opened 2017-04-18
Add pulp_rpmbind type


alexjfisher:
It'd be useful if after using pulp_register, I could declare a number of pulp_rpmbind resources.

PR #231, opened 2017-04-19
Add pulp_rpmbind type

alexjfisher:
To be used on agent (pulp-consumer) nodes to (un)bind to RPM repos.

fixes #GH-230

### DIFF
--- a/lib/puppet/provider/pulp_rpmbind/consumer.rb
+++ b/lib/puppet/provider/pulp_rpmbind/consumer.rb
@@ -1,0 +1,57 @@
+require 'pathname'
+
+Puppet::Type.type(:pulp_rpmbind).provide(:consumer) do
+  desc 'Bind/unbind to an RPM repo'
+
+  confine osfamily: :redhat
+  commands consumer: '/bin/pulp-consumer'
+  commands grep:     '/bin/grep'
+
+  def self.repo_file
+    default = '/etc/yum.repos.d/pulp.repo'
+    begin
+      repo_file = grep('-oP', '^repo_file:\s+\K.*', '/etc/pulp/consumer/consumer.conf')
+      repo_file.strip!
+    rescue Puppet::ExecutionFailure => e
+      Puppet.debug "Couldn't find repo_file configuration in /etc/pulp/consumer/consumer.conf -> #{e.inspect}"
+      return default
+    end
+    return repo_file if (Pathname.new repo_file).absolute?
+    Puppet.debug "repo_file #{repo_file} is not an absolute path.  Using default"
+    default
+  end
+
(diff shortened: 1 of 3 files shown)


lib/puppet/provider/pulp_rpmbind/consumer.rb
```diff
@@ -0,0 +1,41 @@
+Puppet::Type.type(:pulp_rpmbind).provide(:consumer) do
+  desc 'Bind/unbind to an RPM repo'
+
+  confine osfamily: :redhat
+  commands consumer: '/bin/pulp-consumer'
+  commands grep:     '/bin/grep'
+
+  def self.instances
+    begin
+      binds = grep('-oP', '^\[\K.*(?=\]$)', '/etc/yum.repos.d/pulp.repo')
```
ekohl: Is ```pulp.repo``` the only one used by ```pulp-consumer```?

alexjfisher: It appears that way.  I was a bit surprised the pulp-consumer binary couldn't list bindings. Admittedly it does seem hacky to grep the pulp.repo file, but it is very simple to do.  The other option is to use the [API](https://docs.pulpproject.org/dev-guide/integration/rest-api/consumer/bind.html) but that would be *way* more complicated and I can't personally see the advantage.  To the contrary, if someone had manually deleted an entry from pulp.repo, the server wouldn't know about it.

ekohl: I'm leaning to going with grep for now but create an issue for pulp-consumer to list repos.

alexjfisher: So `/etc/yum.repos.d/pulp.repo` is the [default](https://github.com/pulp/pulp/blob/764fe65af86c72996a109a4769e7644b274b29d8/client_consumer/pulp/client/consumer/config.py#L26) in pulp , but it *is* possible to override it in `/etc/pulp/consumer/consumer.conf` and this module does even expose that [option](https://github.com/Katello/puppet-pulp/blob/a48570999513e3f2cae5a4e84a373e836e048850/manifests/consumer/params.pp#L21).

I'm not sure how many people actually do this, (maybe none), but I suppose I should code for it.

alexjfisher: Done!  https://github.com/Katello/puppet-pulp/pull/231/commits/ad4d4a2b5147686f06f18b22fba39baf22742cdc

@ekohl Let me know what you think.  If you're happy with the change, I'll squash the commits.

alexjfisher:
squashed

PR #232, opened 2017-04-20
Remove mongodb_version fact

alexjfisher:
A couple of issues.

1. `lib/facter/facts.rb` is a really bad name for a fact file since it has
to be unique across all modules.
2. It's not even used!  Puppetlabs/mongodb (which is a dependency) provides an identically named
fact that takes precedence (due to [`mongodb_version.rb`](https://github.com/puppetlabs/puppetlabs-mongodb/blob/master/lib/facter/mongodb_version.rb) coming after `facts.rb` alphabetically)

The puppetlabs fact returns a version as reported by mongodb (eg
2.6.12), whereas this module's version would return the package version
(eg 2.6.12-4.el7).  This isn't important as the only place it's used is
in [templates/server.conf.erb](https://github.com/Katello/puppet-pulp/blob/a48570999513e3f2cae5a4e84a373e836e048850/templates/server.conf.erb#L52)


metadata.json
```diff
@@ -22,7 +22,7 @@
     },
     {
       "name": "puppetlabs/mongodb",
-      "version_requirement": ">= 0.10.0 < 1.0.0"
+      "version_requirement": ">= 0.17.0 < 1.0.0"
```
alexjfisher: puppetlabs/mongodb hasn't always provided this fact and may have been too broken before https://github.com/puppetlabs/puppetlabs-mongodb/commit/81ac1d7ff7267255f3de4653bd26c90feb0b829c

ekohl:
On mobile so not linking. This used to be used but when I removed EL6 compat that was no longer the case. Pretty sure you're right that it can be removed

stbenjam:
The puppetlabs-mongodb fact isn't particularly useful if you're installing mongo in the same run as the configuration which is likely on a fresh system.

How does the ordering happen, does it fall back to ours if the mongodb one resolves as nil?

I don't know how common this feature is used, but you could have a problem setting up pulp if you're configuring mongo + pulp in the same puppet run, and the mongo database is being configured with a username/password.  The pulp install will fail because `mongodb_version` will be nil and pulp config will end up without the credentials in the file.

alexjfisher:
@stbenjam You're right.  That's what I missed.  This module's fact resolves *before* mongodb is installed *and* it does then take precedence over the version from puppetlabs/mongodb.  I think this behaviour shouldn't be relied on though.  Perhaps we should just rename the fact to `available_mongodb_version` or something?

alexjfisher:
or, since EL6 isn't supported anymore and epel 7 ships mongodb > 2.6, perhaps remove the fact *and* the [conditional ](https://github.com/Katello/puppet-pulp/blob/a48570999513e3f2cae5a4e84a373e836e048850/templates/server.conf.erb#L52)in the template?

stbenjam:
I think that makes the most sense.  No one should be out there using 2.4 still, and even if they are everything else other than username/password will still work.

spec/spec_helper.rb
```diff
@@ -8,7 +8,6 @@
 
                                                     # Original fact sources:
 add_custom_fact :concat_basedir, '/tmp'             # puppetlabs-concat
-add_custom_fact :mongodb_version, '2.4.14'          # puppetlabs-mongodb
```
ekohl: We should also do this in katello-installer-modulesync.

spec/classes/pulp_admin_spec.rb
```diff
@@ -4,7 +4,7 @@
   context 'on RedHat' do
     context 'install class without parameters' do
       let :facts do
-        on_supported_os['redhat-7-x86_64'].merge(:concat_basedir => '/tmp', :mongodb_version => '2.4.14', :root_home => '/root')
+        on_supported_os['redhat-7-x86_64'].merge(:concat_basedir => '/tmp', :root_home => '/root')
```
ekohl: We can drop the whole merge since we set these facts in ```spec_helper.rb```. Doesn't have to be part of this PR. Just an observation.

metadata.json
ekohl: Given this commit hasn't made it into a release, should we wait for that and depend on it?

alexjfisher: Actually, since adding the commit to remove all use of the fact, I can drop this part of the change.

spec/defines/pulp_apache_fragment_spec.rb
```diff
@@ -5,7 +5,7 @@
 
   context 'on redhat' do
     let :facts do
-      on_supported_os['redhat-7-x86_64'].merge(:concat_basedir => '/tmp', :mongodb_version => '2.4.14', :root_home => '/root')
+      on_supported_os['redhat-7-x86_64'].merge(:concat_basedir => '/tmp', :root_home => '/root')
```
ekohl: The whole merge can be removed here as well.

spec/defines/pulp_child_fragment_spec.rb
```diff
@@ -12,7 +12,7 @@ class {'pulp::child':
 
   context 'on redhat' do
     let :facts do
-      on_supported_os['redhat-7-x86_64'].merge(:concat_basedir => '/tmp', :mongodb_version => '2.4.14', :root_home => '/root')
+      on_supported_os['redhat-7-x86_64'].merge(:concat_basedir => '/tmp', :root_home => '/root')
```
ekohl: The whole merge can be removed here as well.

ekohl:
Thanks!

PR #233, opened 2017-04-20
Add pulp_consumer_id fact

alexjfisher:
It's useful to know if you're already registered with a pulp server.
If not, (as well as including pulp::consumer and using pulp_register),
I want to be able to declare a yumrepo resource for a repo containing
pulp packages.  On the next puppet run, (host is registered), I can
then declare the same resource with ensure => absent and start using
pulp binds instead.

### DIFF
--- a/lib/facter/pulp_consumer_id.rb
+++ b/lib/facter/pulp_consumer_id.rb
@@ -1,0 +1,8 @@
+require 'facter/util/pulp'
+
+Facter.add(:pulp_consumer_id) do
+  confine kernel: 'Linux'
+  setcode do
+    Facter::Util::Pulp.pulp_consumer_id
+  end
+end
--- a/lib/facter/pulp_consumer_server.rb
+++ b/lib/facter/pulp_consumer_server.rb
@@ -1,0 +1,8 @@
+require 'facter/util/pulp'
+
+Facter.add(:pulp_consumer_server) do
+  confine kernel: 'Linux'
+  setcode do
+    Facter::Util::Pulp.pulp_consumer_server
+  end
+end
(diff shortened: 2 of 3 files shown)


lib/facter/pulp_consumer_id.rb
```diff
@@ -0,0 +1,14 @@
+Facter.add(:pulp_consumer_id) do
+  confine kernel: 'Linux'
+  setcode do
+    pulp_consumer_id = nil
+    status = Facter::Util::Resolution.exec('pulp-consumer status')
+    unless status.nil?
+      # Strip color from command output
+      status.gsub!(/\e\[([;\d]+)?m/, '')
+      captures = /^This consumer is registered to the server\s\[(.*)\]\swith\sthe\sID\s\[(.*)\]\.$/.match(status)
```
alexjfisher: regex uses `\s` as the command output is automatically wrapped using this [library](http://okaara.readthedocs.io/en/latest/prompt-api.html#prompt-class-apis).

stbenjam: Do you want to do something with match[0] - like register an additional fact for the pulp server?

alexjfisher: Could do; perhaps it was a bit lazy of me not to. (Personally, I'm likely to only use the fact as if it were a boolean).
I just need to think about the correct way and style.
What is the preferred style?  I *could* create a structured fact??   If not, should each fact be in it's own file?  Does code shared between facts belong in it's own namespace? That sort of thing.  I'll figure it out.  Pretty sure I shouldn't just nest `Facter.add` blocks anyway!

alexjfisher: More importantly, the current code is wrong and is already returning the server_id not consumer_id.  Let me have another go...

alexjfisher:
@ekohl Code wasn't correct.  Oops!  Should be now.  Would you like to re-review?

ekohl:
I shouldn't review code just after coming back from vacation.

stbenjam:
Thanks @alexjfisher!

PR #234, opened 2017-04-21
Add basic provider for Pulp roles

ccnifo:
Thanks to crfistifalcas various providers and the help of riton, here is a provider for Pulp roles.

It adds/removes roles, ensures the desired users are members of it and manages hashes representing URLs and associated permissions. It does not manage the "display name" and "description" attributes, though.

### DIFF
--- a/lib/puppet/provider/pulp_role/api.rb
+++ b/lib/puppet/provider/pulp_role/api.rb
@@ -1,0 +1,113 @@
+
+require File.expand_path('../../../util/pulp_util', __FILE__)
+
+Puppet::Type.type(:pulp_role).provide(:api) do
+  commands :pulp_admin => '/usr/bin/pulp-admin'
+
+  mk_resource_methods
+
+  def initialize(resource={})
+    super(resource)
+    @pulp = Puppet::Util::PulpUtil.new
+  end
+
+  def exists?
+    r = @pulp.get_role_info(name())
+    !r.nil?()
+  end
+
+  def create
+    manage_role('create', name())
+  end
+
+  def destroy
+    manage_role('delete', name())
+  end
+
+  def users
+    users = @pulp.get_role_users(name())
+    if users.nil?
+      nil
+    end
+
+    @users = users
+    @users
+  end
+
+  def users=(should)
+    users_to_add = should - @users
+    users_to_rm = @users - should
+    users_to_add.each() do |user|
+      if user_exist(user)
+        role_user_mod("add", user, name())
+      else
+        Puppet::Util::Warnings.warnonce("#{user} does not exist in Pulp, can't be added to any role")
+      end
+    end
+    users_to_rm.each() do |user|
+      role_user_mod("remove", user, name())
+    end
+  end
+
+  def permissions
+    permissions = @pulp.get_role_permissions(name())
+    if permissions.nil?
+      nil
+    end
+
+    @permissions = permissions
+    @permissions
+  end
+
+  def permissions=(should)
+    should.each_key() do |k|
+      # if key is already there, inspect each permission associated
+      if @permissions.has_key?(k)
+        if should[k] != @permissions[k]
+          perms_to_add = should[k] - @permissions[k]
+          perms_to_add.each() do |perm|
+            role_permission_mod("grant", k, perm, name())
+          end
+
+          perms_to_remove = @permissions[k] - should[k]
+          perms_to_remove.each() do |perm|
+            role_permission_mod("revoke", k, perm, name())
+          end
+        end
+      # if key is absent, add all permissions
+      else
+        should[k].each() do |perm|
+          role_permission_mod("grant", k, perm, name())
+        end
+      end
+    end
+
+    # Remove a key which should not be there anymore
+    @permissions.each_key() do |k|
+      if ! should.has_key?(k)
+        @permissions[k].each() do |perm|
+          role_permission_mod("revoke", k, perm, name())
+        end
+      end
+    end
+
+  end
+
+  def user_exist(login)
+    stdout = pulp_admin(['auth', 'user', 'search', '--str-eq', 'login='+login])
+    return true if stdout.to_s =~ /Login:/
+  end
+
+  def manage_role(action, role_id)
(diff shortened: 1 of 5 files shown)


ccnifo:
Maybe Pulp roles are not a very popular functionality? The changes have been tested and work fine on my Pulp setup, but if you have clues to test further I would be glad to read about that.

ccnifo:
I added a function to check existence of every user to add to a role. It displays a warning if the login doesn't exist, rather than creating an error.

ccnifo:
I forgot to push this condition, which just aims at distinguishing a normal "not found" API answer from a 500 error.

ccnifo:
I read a release is planned soon: do you think it should include the roles provider?

ekohl:
I'm currently working on refactoring the existing types and providers, including writing tests. There I also ran into the issue with the 404 vs error. That caused https://github.com/Katello/puppet-pulp/issues/245 so 👍 on that. I made a similar change before I reviewed yours.

Regarding a release: I hadn't planned on it yet. The only thing is that we know it's going to be a major release since we're using puppet 4 types. That means we can break some things (like changing the default of enabling oauth).

lib/puppet/type/pulp_role.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  newparam(:name, :namevar => true) do
+    desc "role-id: uniquely identifies the role"
+  end
+
+  newproperty(:users, :array_matching => :all) do
+    desc "users to add to this role"
+  end
+
+  newproperty(:permissions) do
```
ekohl: This is also an array. Should it match all like users?

ccnifo: No, it's a hash which contains arrays. It's thus handled in a much more verbose way (I couldn't make it shorter) in lib/puppet/provider/pulp_role/api.rb

ekohl:
My effort to clean up the types is in https://github.com/Katello/puppet-pulp/pull/252 btw. I expect to be able to do some testing of this soon based on what I've learned in cleaning up.

ekohl:
I merged my type cleanup patch and this needs a rebase now. If you don't have time I could take a look at it.

ccnifo:
OK, I worked on this today: nice work, I just got a small problem with pulp_schedule. I'll report the issue and get back to this PR when it's resolved.

ccnifo:
OK, update done. @ekohl: based on your work, I also added an example manifest.

lib/puppet/provider/pulp_role/api.rb
```diff
@@ -0,0 +1,113 @@
...
+  end
+
+  def destroy
+    manage_role('delete', name())
+  end
+
+  def users
+    users = @pulp.get_role_users(name())
+    if users.nil?
+      nil
```
ekohl: Should this be a ```return nil```?

lib/puppet/provider/pulp_role/api.rb
```diff
@@ -0,0 +1,113 @@
...
+    end
+    users_to_rm.each() do |user|
+      role_user_mod("remove", user, name())
+    end
+  end
+
+  def permissions
+    permissions = @pulp.get_role_permissions(name())
+    if permissions.nil?
+      nil
```
ekohl: Should this also be ```return nil```?

lib/puppet/provider/pulp_role/api.rb
```diff
@@ -0,0 +1,113 @@
...
+      else
+        should[k].each() do |perm|
+          role_permission_mod("grant", k, perm, name())
+        end
+      end
+    end
+
+    # Remove a key which should not be there anymore
+    @permissions.each_key() do |k|
+      if ! should.has_key?(k)
```
ekohl: Convention is to use ```unless``` instead of ```if !```

lib/puppet/util/pulp_util.rb
```diff
@@ -50,6 +50,23 @@ def get_repo_info(repo_id)
         request_api("/v2/repositories/#{repo_id}/?details=True")
       end
 
+      def get_role_info(role_id)
+        raise '[get_role_info] Role id should never be nil.' if !role_id
```
ekohl: Convention is to use ```unless``` instead of ```if !```. This also applies to the lines below.

lib/puppet/provider/pulp_role/api.rb
ccnifo: Yes, I'm not used to Ruby at all, but I thought ``nil`` ``return nil`` and ``return`` were the same thing? Is ``return nil`` a better style?
I just tried roles with empty user array and empty permissions hash, it works well with both syntaxes.

lib/puppet/provider/pulp_role/api.rb
ccnifo: Corrected in 31da5bc , thanks

lib/puppet/util/pulp_util.rb
ccnifo: Corrected in 31da5bc , thanks

ccnifo:
OK, thanks for the review and sorry for the mess when I updated my branch and accidentally closed the PR.

lib/puppet/provider/pulp_role/api.rb
ccnifo: Yes, same answer as above

lib/puppet/provider/pulp_role/api.rb
ekohl: The difference is that ```return nil``` results in ```@users``` not being set. Both return nil in the end. In the current code this if-block does nothing so it can either be removed or changed to ```return nil```. Not sure which is the better way.

ccnifo: You are right, those lines are of no use. I deleted them in cef1604 , thanks!

lib/puppet/provider/pulp_role/api.rb
ccnifo: deleted in cef1604.

ekohl:
Thanks for sticking with it. I know it took a long time.

PR #237, opened 2017-05-10
Use the $pulp::ca_cert variable rather than hardcoding

ekohl:
The pulp-gen-ca-certificate reads the location from the config file, so
the output location is variable.


jlsherrill:
ACK thanks @ekohl !

PR #239, opened 2017-05-11
Refactor the class inclusions and chaining

ekohl:


ekohl:
@cristifalcas and @alexjfisher: I know you run this module outside of the installer context. Would these changes work in your environment?

Note I'm still testing this against our installer pipeline, but I'd like some feedback.

alexjfisher:
@ekohl I'd be happy to test against my pulp server role.  Let me know when you're ready.

ekohl:
@alexjfisher awesome, will do!

alexjfisher:
For what it's worth, there's no class containment in [consumer.pp](https://github.com/Katello/puppet-pulp/blob/3bf701bacc8d98cdb49ac37d301322fbce941163/manifests/consumer.pp#L135)

ekohl:
The current problem I'm running into is that ```pulp-server``` installs ```/etc/httpd/conf.d/pulp.conf```, then Apache is restarted because of ```Concat[28-katello-reverse-proxy.conf]: Scheduling refresh of Class[Apache::Service]``` and results into:

```
[ERROR 2017-05-12 12:17:21 verbose] May 12 12:17:21 pipeline-proxy-nightly-centos7.example.com httpd[27117]: AH00526: Syntax error on line 21 of /etc/httpd/conf.d/pulp.conf:
[ERROR 2017-05-12 12:17:21 verbose] May 12 12:17:21 pipeline-proxy-nightly-centos7.example.com httpd[27117]: SSLCACertificateFile: file '/etc/pki/pulp/ca.crt' does not exist or is empty
[ERROR 2017-05-12 12:17:21 verbose] May 12 12:17:21 pipeline-proxy-nightly-centos7.example.com systemd[1]: httpd.service: main process exited, code=exited, status=1/FAILURE
```

ekohl:
I finally figured out why it won't work on the first run, but does on the second and nothing relevant in the puppet logs! https://github.com/Katello/katello-installer/blob/f05e8d737e4a463c16bb3694c61e8c6e13ebd359/hooks/post/29-create_package_httpd_conf.rb wipes the file but that happens after Apache is supposed to be started. It contains a reference to a file that does not exist.

ekohl:
The installer now passes, commits have been cleaned up so I think it's ready for proper review.

manifests/apache.pp
```diff
@@ -107,13 +107,24 @@
       # allow older yum clients to connect, see bz 647828
       custom_fragment            => 'SSLInsecureRenegotiation On',
     }
+
+    # This file is installed by pulp-server but we have everything in the above vhost
+    file {'/etc/httpd/conf.d/pulp.conf':
+      ensure  => file,
+      content => '# This file is managed by the foreman-installer, do not alter.\n',
```
stbenjam: Nitpick, but maybe say managed by the puppet-pulp module in case of running outside of foreman-installer

ekohl: Initially I just copied this from the installer, but even that differs slightly because that has no line ending.

alexjfisher:
@ekohl I'm going to test this branch out now.  Will report back shortly...

manifests/apache.pp
alexjfisher: With single quotes, this has no line ending either. ;)

ekohl:
Updated with a new message and I've added a line to the spec test.

ekohl:
Thanks for testing guys

Issue #241, opened 2017-05-18
How to use Pulp - need docs


djjudas21:
I'm new to Pulp but I installed it with the Puppet module `katello/pulp` using this very basic manifest to begin with:

```puppet
class profile::pulp {
  include ::profile::apache
  include ::profile::apache::ssl
  include ::apache::mod::wsgi

  yumrepo { 'pulp-2-stable':
    descr               => 'Pulp 2 Production Releases',
    baseurl             => 'https://repos.fedorapeople.org/repos/pulp/pulp/stable/2/$releasever/$basearch/',
    enabled             => 1,
    skip_if_unavailable => 1,
    gpgcheck            => 1,
    gpgkey              => 'https://repos.fedorapeople.org/repos/pulp/pulp/GPG-RPM-KEY-pulp-2',
  }

  # katello/pulp
  class { '::pulp':
    enable_rpm    => true,
    enable_puppet => true,
    require       => Yumrepo['pulp-2-stable'],
  }
}
```

This runs cleanly on my CentOS 7 node, seems to install all the correct dependencies etc. I've followed the setup steps in the [Pulp docs](http://docs.pulpproject.org/en/2.13/user-guide/installation/f23-.html) and each step seems to be completed OK. However the Pulp application is non-functional. Navigating to `http://pulp.example.com` just prints out a directory listing, mostly `.wsgi` files which simply display the content if I click on them. Navigating to `https://pulp.example.com` gives Error 403 (Forbidden). There is nothing helpful in the Apache logs. I don't know much about Python/wsgi so I'm stuck.

So **either** the `katello/pulp` module is broken and doesn't install Pulp properly, **or** there is insufficient documentation to enable beginners to actually the module. Judging by the scores on [Puppet Forge](https://forge.puppet.com/katello/pulp/scores), almost everyone has rated the module as being too hard to use and having insufficient docs. I've read the comments in the manifests and they give a reasonable idea of what each individual parameter does, but the docs are desperately missing an overall guide of which components to install and how to do it.

For example, look at the now-deprecated [`pulp/pulp`](https://forge.puppet.com/pulp/pulp) module, which has explanatory notes and a complete worked example. Please would it be possible to add something similar to `katello/pulp`? I think it would benefit a *lot* of people. Thanks.

ekohl:
[...]

Issue #143, opened 2016-04-19
Update documentation with all settings available


rammaram06:
Hi,

If possible, could you guys please update the documentation for this module with different settings it provides and how to use them? Something similar to old pulp-pulp module.

Thanks

ehelms:
What is the old pulp-pulp module that you mention? And what sort of documentation are you wanting? All parameters are documented at the top of the manifests, e.g. https://github.com/Katello/puppet-pulp/blob/master/manifests/init.pp#L5

ramreddy26:
Something like this? https://forge.puppet.com/pulp/pulp

stbenjam:
Generally we don't put this in the README as it gets stale fast, the top of the .pp files documents the parameters.

For all the classes we intended users to use directly, we even have a test uses puppet lint to ensure every single param is documented.

lmayorga1980:
Wondering if there is a parsing class documentation tags to make them part of the README.md.

rgarrigue:
Hi

Just a :+1: about this topic, documentation is really lacking until you get the idea to have a look directly in the .pp

And if I'm not mistaken this module is the successor of pulp/pulp. You should update pulp/pulp README.md to link to this module, and set version to 9.999.999 (see the nginx modules in forge)

Best regards,

ekohl:
Something like http://www.puppetmodule.info/modules/katello-pulp/puppet_classes/pulp then?

rgarrigue:
Not really. I was thinking about something like
https://forge.puppet.com/puppet/nginx, a handfull of simple case examples
to display the possibilities, something that will be unlikely to deprecate
much. Then maybe a big complex example for the in-deep. Couple of comments
in between to get started. And a link in the end to you full reference
above :-)

2017-03-08 8:45 GMT+01:00 Ewoud Kohl van Wijngaarden <
notifications@github.com>:

ekohl:
That makes total sense and it's feedback that I received at cfgmgmtcamp as well about theforeman modules. I'd be very interested in getting some examples of what users expect of it.

ekohl:
With the updated docs and http://www.puppetmodule.info/modules/katello-pulp/puppet_classes/pulp, is this sufficiently addressed?

rgarrigue:
@ekohl Seems very good. One suggestion though, add this puppetmodule.info link to the README.md (I serached quickly, maybe I missed it if it's here)

ekohl:
@rgarrigue it's not and I will add it.

Issue #164, opened 2016-09-08
EPEL and Pulp repos not deployed


ccnifo:
pulp::install relies on yum to install pulp RPMs, but the repos.fedorapeople.org and EPEL repos mentioned in the Pulp doc are not deployed.
I suppose managing them in this module would be logical, ideally with a switch to make it optional. Do you agree? If yes, do you have any comments regarding how to implement it?

riton:
BTW, current `pulp` main class design does not protect against resource containment and something like

```puppet
class { '::setup_my_repositories': } ->
class { '::pulp': }
```

leads to a _chicken and egg_ problem.
_pulp_ packages will try to install before any repository defined in the `setup_my_repositories` class get actually available.

Would you mind using _anchors_ / _contain_ ?
This would solve this problem as discussed [here](https://www.devco.net/archives/2012/12/13/simple-puppet-module-structure-redux.php#comment-175995) ?

rgarrigue:
I got issues with repo and managed to get it working like this

```
Yumrepo <| |> ~> Exec['yum-cache'] -> Package <| |>

    # Solves conflict between compat-qpid-* et qpid-*
    yumrepo { 'epel':
      exclude => 'compat-qpid*',
    }
    # From https://repos.fedorapeople.org/repos/pulp/pulp/rhel-pulp.repo
    yumrepo { 'pulp-2-stable':
      baseurl             => 'https://repos.fedorapeople.org/repos/pulp/pulp/stable/2/$releasever/$basearch/',
      descr               => 'Pulp 2 Production Releases',
      enabled             => true,
      gpgcheck            => true,
      gpgkey              => 'https://repos.fedorapeople.org/repos/pulp/pulp/GPG-RPM-KEY-pulp-2',
      skip_if_unavailable => true,
    }
    ~>
    exec { 'yum-cache':
      command => '/usr/bin/yum clean all; /usr/bin/yum makecache',
    }
    ->
    class { '::pulp': ........ }
```

ekohl:
In our installer we rely on other modules to do this. EPEL is already needed by our installer and otherwise puppet-foreman does this for us. The pulp repos are installed by katello-release. I'm not sure what the best solution here is.

ekohl:
[...]

PR #242, opened 2017-05-21
Update the README

ekohl:
Still not very good, but an improvement aimed to address https://github.com/Katello/puppet-pulp/issues/143 and https://github.com/Katello/puppet-pulp/issues/241.

@rammaram06 @rgarrigue @djjudas21 if you could have a look then that'd great.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,6 @@
 [![Puppet Forge](http://img.shields.io/puppetforge/v/katello/pulp.svg)](https://forge.puppetlabs.com/katello/pulp)
 [![Build Status](https://travis-ci.org/Katello/puppet-pulp.svg?branch=master)](https://travis-ci.org/Katello/puppet-pulp)
+
 #### Table of Contents
 
 1. [Overview](#overview)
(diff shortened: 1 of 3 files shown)


rammaram06:
@ekohl this looks good..would you be able to add an example on how to setup ssl configuration?

ekohl:
I've added some SSL examples, but haven't tested it yet. Looks like I'm still missing some of the settings on the messaging.

djjudas21:
This looks much more useful. I'll try it out properly tomorrow - thanks!

rgarrigue:
Some suggestions

- There's a debug parameter, document it and explain where's the output if
it's a specific file (as opposed to puppet output)
- Document emails settings (email_enabled email_from email_host email_port)
- In issue #164 we mention pulp rpm repo management, as far as I'm
concerned it's mandatory for pulp standalone installation (w/o katello), it
ought to be mentionned too
- default_login and default_password can be mentionned, with the default
value (admin / random pwd)
- Add a quick exemple to open firewall port with crayshfix's firewalld
module, since it's all about centos 7

That's it for what I'm doing. Reading the params.pp, an example with a
proxy configuration might interest people too

Best regards,


2017-05-22 17:25 GMT+02:00 Jonathan <notifications@github.com>:

ekohl:
I'm trying to simplify the usage of the admin in https://github.com/Katello/puppet-pulp/pull/248. Once that's in I'm going to expand the README.

ekohl:
Updated.

> - There's a debug parameter, document it and explain where's the output if it's a specific file (as opposed to puppet output)

I'm not sure what these exactly do so I've left them out for now.

> - Add a quick exemple to open firewall port with crayshfix's firewalld module, since it's all about centos 7

Personally I've always used the puppetlabs-firewall module so I'm leaving this out for now. I could mention the ports used.

> That's it for what I'm doing. Reading the params.pp, an example with a proxy configuration might interest people too

I noticed we have the parameter ```proxy_url``` which maps to ```proxy_host``` in the config. @ehelms you added this in 6388c094f735c6a59d2ab3440d5e8b6f8051cfc9 so can you have a look?

ehelms:
@ekohl guessing it was a bit of a typo when configuring this for use with Katello's installer

ekohl:
Given we're going to do a major version, shall we change it to ```proxy_host```?

ekohl:
I'd like to get his merged on Tuesday because even if this is still incomplete, I think it's better than what we have now. Please point out any mistakes though. Suggestions for additions are also welcome and will be addressed in a separate PR.

Issue #243, opened 2017-05-23
Facts broken with some ruby versions


ccnifo:
Using the up-to-date module code today, facts broke with this message:

> Error loading fact /var/lib/puppet/lib/facter/pulp_consumer_server.rb: /var/lib/puppet/lib/facter/pulp_consumer_server.rb:4: syntax error, unexpected ':', expecting kEND
>   confine kernel: 'Linux'
>                  ^
> Error loading fact /var/lib/puppet/lib/facter/pulp_consumer_server.rb: /var/lib/puppet/lib/facter/pulp_consumer_server.rb:4: syntax error, unexpected ':', expecting kEND
>   confine kernel: 'Linux'
>                  ^
> Error loading fact /var/lib/puppet/lib/facter/pulp_consumer_id.rb: /var/lib/puppet/lib/facter/pulp_consumer_id.rb:4: syntax error, unexpected ':', expecting kEND
>   confine kernel: 'Linux'
>                  ^
> Error loading fact /var/lib/puppet/lib/facter/pulp_consumer_id.rb: /var/lib/puppet/lib/facter/pulp_consumer_id.rb:4: syntax error, unexpected ':', expecting kEND
>   confine kernel: 'Linux'
>                  ^

A quick search makes me think this is due to ruby 1.8 not supporting this new syntax, rather expecting something like _confine :kernel => 'Linux'_. Do you confirm?

ccnifo:
I forgot to mention that on a machine running ruby 2.0, the issue does not appear.

stbenjam:
We dropped Ruby 1.8.7 support in puppet-pulp 4.0.  You can use an earlier version if you need it, 3.5 was the last release we supported Ruby 1.8.7 and CentOS 6 IIRC.

https://forge.puppet.com/katello/pulp/3.5.0/

ekohl:
We no longer test with those ruby versions since we officially dropped support for EL6 so that's entirely possible. I assume this happens because you have some EL6 clients in your environment where this puppet module is used as well. I think we could accept a patch for this.

ccnifo:
You're right the problem appears on EL6 clients, but the thing is that they are not related to the module: they just share the same environment, which means facts are deployed on them as well.
I thus have to write a patch for our own needs; I'll open a PR so that you can choose to merge or not.

stbenjam:
[...]

PR #246, opened 2017-05-24
Fixes GH-243 - Ruby 1.8 compatible syntax for facts

ccnifo:
These are just syntax changes to maintain facts compatibility with Ruby 1.8.7.
These changes aim at allowing EL6 systems to share the same Puppet environment as Pulp servers, without breaking facts collection on EL6.


ekohl:
Thanks!

Issue #138, opened 2016-03-31
Pulp v2.8 does not work with SSL_CLIENT_S_DN_CN 


aeria:
When using this pulp module by itself, after installation pulp-admin won't work for most commands.

/var/log/messages shows us:
`Mar 31 00:18:22 localhost pulp: pulp.server.webservices.middleware.exception:INFO: Authentication with username admin:admin:56fc5e8c5af5cd2789b2cdfc failed: preauthenticated remote user is missing.`

The apache log shows:
`logs/pulp-https_error_ssl.log:[Wed Mar 30 23:23:18.891305 2016] [ssl:warn] [pid 10862] [client 127.0.0.1:33780] AH02227: Failed to set r->user to 'SSL_CLIENT_S_DN_CN'
`

Removing this line made it all work again:
`conf.d/05-pulp-https.conf:SSLUsername SSL_CLIENT_S_DN_CN`

Someone with the same issue on a mailing list:
http://comments.gmane.org/gmane.linux.redhat.fedora.pulp/4099

ehelms:
Hi Taz,

Can you give more info on your setup and configuration to help us reproduce?

aeria:
Absolutely

`puppet module install katello-pulp`

  include ::apache
  include ::apache::mod::proxy
  include ::apache::mod::proxy_http
  include ::epel

And in my .pp file

`  class {'::pulp':
    enable_rpm    => true,
    enable_http   => true,
    manage_httpd  => true,
  }

  class {'::pulp::admin':
    enable_rpm => true,
    verify_ssl => false,
    require       => Class['::pulp'],
  }
`

I noticed this module doesn't have any acceptance tests yet. I'm looking at potentially writing some, but shying away as this Puppet module is not idempotent. Every run 'restarts' the workers daemon, which by its nature is never actually running but forks off.

jlambert121:
I ran into the same issue this weekend with pulp 2.8.2, same fix.

ehelms:
The inclusion of SSL_CLIENT_S_DN_CN is what allows for using certificate authorization to talk to Pulp vs using oauth. I think this might be more of a Pulp issue where it should be able to handle this situation . We could conditionally set this directive in the Apache configuration but I think ultimately Pulp should be able to handle it.

jlambert121:
[...]

PR #248, opened 2017-05-28
pulp-admin refactoring

ekohl:
This refactoring relies on not having a SSLUsername (https://github.com/Katello/puppet-pulp/issues/138) so it's broken in the default deployment. You can create a /root/.pulp/admin.conf with the credentials hardcoded there as well. Would that be preferable?

It also looks like the providers need to be modified to get more autorequires for correct ordering.

### DIFF
--- a/.sync.yml
+++ b/.sync.yml
@@ -1,4 +1,7 @@
 ---
+.travis.yml:
+  beaker_sets:
+    - docker/centos-7
 Rakefile:
   param_docs_pattern:
     - manifests/admin.pp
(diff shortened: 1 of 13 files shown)


ekohl:
@ehelms could you have a look? I'd like to fix up the providers to be able to use HTTP basic auth so we can remove the warning but this is the best for now I think.

ekohl:
I included the result of https://github.com/Katello/foreman-installer-modulesync/pull/28

ekohl:
Includes https://github.com/Katello/puppet-pulp/pull/255

ekohl:
Included in https://github.com/Katello/puppet-pulp/issues/252

ekohl:
I expected the force push to show up here again in the hope to merge the admin parts of #252 but currently it looks like I broke GH and it's not showing up. Can't close the issue either.

PR #251, opened 2017-06-01
fixes #19740 - pulp_docker.conf schema 2

thomasmckay:
correctly handle schema 2

To test, sync solr from docker hub then docker pull w/ docker-1.12 or higher


ekohl:
Odd that the tests appear to pass here but now fail in master.

Issue #245, opened 2017-05-24
unable to add pulp_rpmrepo (undefined method `each' for nil:NilClass)


TJM:
We created our pulp server today and tried to use the pulp_rpmrepo provider to configure it, but it is getting the following error: (end of puppet agent -t --debug)

```
Debug: Prefetching api resources for pulp_rpmrepo
Debug: Storing state
Debug: Stored state in 0.05 seconds
Error: Failed to apply catalog: undefined method `each' for nil:NilClass
Debug: Dynamically-bound server lookup failed, falling back to report_server setting
Debug: Dynamically-bound port lookup failed; falling back to report_port setting
Debug: Using cached connection for https://mbi-foreman-01.utmb.lan:8140
Debug: Caching connection for https://mbi-foreman-01.utmb.lan:8140
Debug: Closing connection for https://mbi-foreman-01.utmb.lan:8140
```

Here is the relevant data:

```
---
pulp::debugging_mode: true
pulp::log_level: DEBUG
pulp::ssl_verify_client: optional
pulp::https_cert: /etc/puppetlabs/puppet/ssl/certs/%{::fqdn}.pem
pulp::https_key: /etc/puppetlabs/puppet/ssl/private_keys/%{::fqdn}.pem
pulp::https_chain: /etc/puppetlabs/puppet/ssl/certs/ca.pem
pulp::admin::verify_ssl: false
pulp::admin::logging_filename: /var/log/pulp/admin.log
pulp::admin::logging_call_log_filename: /var/log/pulp/server_calls.log

profile::pulp::rpmrepos:
  centos-7-x86_64-base:
    display_name: CentOS 7 x86_64 Base
    feed: http://mirror.centos.org/centos/7/os/x86_64/
    relative_url: centos/7/os/x86_64/
    remove_missing: true
    retain_old_count: 2
    serve_http: true
    serve_https: true
    validate: true
  centos-7-x86_64-updates:
    display_name: CentOS 7 x86_64 Updates
    feed: http://mirror.centos.org/centos/7/updates/x86_64/
    relative_url: centos/7/updates/x86_64/
    remove_missing: true
    retain_old_count: 2
    serve_http: true
    serve_https: true
    validate: true
  centos-7-x86_64-epel:
    display_name: CentOS 7 x86_64 EPEL
    feed: http://download.fedoraproject.org/pub/epel/7/x86_64/
    relative_url: epel/7/x86_64/
    remove_missing: true
    retain_old_count: 2
[...]

Issue #205, opened 2017-03-09
Idempotent repo creation


rgarrigue:
Hello

I'm ending up with an Error on a second run whenever a repository is already created.

```
==> test.vagrant.local: Error: /Stage[main]/Profiles::Pulp/Pulp_puppetrepo[puppetlabs]: Could not evaluate: Execution of '/usr/bin/pulp-admin puppet repo create --repo-id puppetlabs --display-name Forge Puppetlabs --feed http://forge.puppetlabs.com/ --validate false --verify-feed-ssl false --max-downloads 10 --serve-http true --serve-https false' returned 65: A resource with the ID "puppetlabs" already exists.
```

We've looked around a bit in api.rb, there's no existence check 

Regards,

cristifalcas:
There is no need for existence check with puppet providers.

All existing resources are created by the provider when the puppet run starts. Puppet knows about declared resources inside manifests and it decides if a resource already exists or it should be created.

The error is very strange, because from my point of view, it should not happen. 

Can you post the output from this:

    pulp-admin puppet repo list --details --repo-id puppetlabs
    puppet resource pulp_puppetrepo

And the definition from the manifest of pulp_puppetrepo[puppetlabs]

rgarrigue:
Here's my first run

```
==> test.vagrant.local: Error: Could not prefetch pulp_rpmrepo provider 'api': https request for repo "# threw exception SSL_connect returned=1 errno=0 state=SSLv3 read server certificate B: certificate verify failed. Connection details: url=https://test.vagrant.local:443/pulp/api/v2/repositories///, cert_path=/root/.pulp/user-cert.pem.
==> test.vagrant.local: Notice: /Stage[main]/Profiles::Pulp/Pulp_rpmrepo[epel-7]/ensure: created
==> test.vagrant.local: Error: /Stage[main]/Profiles::Pulp/Pulp_rpmrepo[epel-7]: Could not evaluate: https request for repo "epel-7# threw exception SSL_connect returned=1 errno=0 state=SSLv3 read server certificate B: certificate verify failed. Connection details: url=https://test.vagrant.local:443/pulp/api/v2/repositories/epel-7///?details=True, cert_path=/root/.pulp/user-cert.pem.
[...]

PR #252, opened 2017-06-06
Clean up types and providers

ekohl:
Closes GH-205
Closes GH-245

### DIFF
--- a/examples/pulp_isorepo.pp
+++ b/examples/pulp_isorepo.pp
@@ -1,0 +1,29 @@
+class { '::pulp::repo::upstream': } ->
+class { '::pulp':
+  # https://github.com/Katello/puppet-pulp/issues/138
+  ssl_username => '',
+  enable_admin => true,
+}
+
+pulp_isorepo { 'DataMonster':
+  display_name    => 'Data Monster',
+  description     => "Cookie's cousin",
+  note            => {'a' => 'b'},
+  feed            => 'http://example.com/feed',
+  validate        => true,
+  feed_ca_cert    => '/etc/pki/pulp/ca.crt',
+  verify_feed_ssl => true,
+  feed_cert       => '/etc/pki/pulp/rsa_pub.key',
+  feed_key        => '/etc/pki/pulp/rsa.key',
+  proxy_host      => 'proxy.example.com',
+  proxy_port      => 8080,
+  # Proxy passwords are never idempotent since we can't retrieve the
+  # password from the API. user requires a password
+  #proxy_user      => 'puser',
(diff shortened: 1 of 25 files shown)


ekohl:
One of the issues is now that the ```pulp_puppetrepo``` adds a plugin which means the DB needs to be updated. That requires all services to be stopped which we currently don't handle.

ekohl:
Includes https://github.com/Katello/puppet-pulp/pull/248

lib/puppet/provider/pulp_rpmrepo/api.rb
```diff
@@ -1,10 +1,18 @@
-require File.expand_path('../../../util/pulp_util', __FILE__)
+require File.expand_path('../../../util/repo_provider', __FILE__)
 
-Puppet::Type.type(:pulp_rpmrepo).provide(:api) do
+Puppet::Type.type(:pulp_rpmrepo).provide(:api, :parent => PuppetX::Pulp::RepoProvider) do
   commands :pulp_admin => '/usr/bin/pulp-admin'
 
   mk_resource_methods
 
+  def self.repo_type
```
ekohl: I think that this is very ugly, but I don't know enough about ruby inheritance to do this cleaner.

ekohl: (you can see my attempt at the last commit)

examples/pulp_isorepo.pp
```diff
@@ -0,0 +1,29 @@
+class { '::pulp::repo::upstream': } ->
+class { '::pulp':
+  # https://github.com/Katello/puppet-pulp/issues/138
+  ssl_username => '',
+  enable_admin => true,
+}
+
+pulp_isorepo { 'DataMonster':
+  display_name    => 'Data Monster',
+  description     => "Cookie's cousin",
```
ehelms: I chuckled good.

ehelms:
I don't see any major issues here, it still bothers me a bit this has to re-implement a Ruby API to talk to Pulp when we have Runcible. But if tests are passing I'm OK with all of this.

ekohl:
Rebased on master, I think this is good for a final review and to be merged now.

ekohl:
So I rebased this on master and the dependency on the unreleased qpid breaks acceptance tests. Does this module really need the unreleased version (and thus work around it by including a git checkout) or should we decrease the minimal version to one that's on the forge?

ekohl:
https://github.com/Katello/puppet-pulp/pull/258 would unblock this.

ekohl:
Looks like I screwed up some rebases and lost some work.

PR #254, opened 2017-06-09
allow configuration of wsgi_processes and wsgi_max_requests

SimonPe:
We are having some issues with pulp's wsgi threads leaking memory, whether this is some low level python issue, something in django, or a bug in pulp itsef. The easiest way to mitigate it is just setting the "maximum-requests" parameter for the wsgi daemon¹.

Since I was adding a wsgi daemon parameter anyhow, I also decided to add a wsgi_processes option, since that looks like something people might want to tweak too.

¹ http://modwsgi.readthedocs.io/en/develop/configuration-directives/WSGIDaemonProcess.html

### DIFF
--- a/manifests/apache.pp
+++ b/manifests/apache.pp
@@ -89,7 +89,14 @@ class pulp::apache {
       ssl_verify_depth           => '3',
       wsgi_process_group         => 'pulp',
       wsgi_application_group     => 'pulp',
-      wsgi_daemon_process        => 'pulp user=apache group=apache processes=3 display-name=%{GROUP}',
+      wsgi_daemon_process        => join([
+          'pulp',
+          'user=apache',
+          'group=apache',
+          "processes=${::pulp::wsgi_processes}",
+          "maximum-requests=${::pulp::wsgi_max_requests}",
+          'display-name=%{GROUP}',
+      ], ' '),
       wsgi_pass_authorization    => 'On',
       wsgi_import_script         => '/usr/share/pulp/wsgi/webservices.wsgi',
       wsgi_import_script_options => {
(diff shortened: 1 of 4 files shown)


manifests/apache.pp
```diff
@@ -89,7 +89,14 @@
       ssl_verify_depth           => '3',
       wsgi_process_group         => 'pulp',
       wsgi_application_group     => 'pulp',
-      wsgi_daemon_process        => 'pulp user=apache group=apache processes=3 display-name=%{GROUP}',
+      wsgi_daemon_process        => join([
+          'pulp',
+          'user=apache',
+          'group=apache',
+          "processes=${pulp::wsgi_processes}",
```
ekohl: We try to use ```$::pulp``` everywhere.

manifests/init.pp
```diff
@@ -232,6 +232,11 @@
 #
 # $max_keep_alive::             Configuration value for apache MaxKeepAliveRequests
 #
+# $wsgi_processes::             Number of WSGI processes to spown for pulp itself
+#
+# $wsgi_max_requests::          Maximum number of requests for each wsgi worker to process before
+#                               shutting down and restarting, usefull to combat memery leaks.
```
ekohl: s/usefull/useful/
s/memery/memory/

ekohl:
Also looks like you need to adjust the tests. https://github.com/Katello/puppet-pulp/pull/251 has introduced one error, the other is the wsgi one.

ekohl:
https://github.com/Katello/puppet-pulp/pull/255 should fix the existing test failure.

SimonPe:
@ekohl fixed those little things now :-)

ekohl:
Thanks! I fixed one last typo in the docs while squashing.

PR #256, opened 2017-06-16
Bump qpid dependency

ehelms:


sean797:
ping @ehelms, can you publish that newer qpid version on puppet forge please?

Issue #260, opened 2017-07-13
pulp_schedule broken


ccnifo:
Working on the freshly released 5.0 version, I got an issue with the pulp_schedules stored in my config:
`Error: Could not prefetch pulp_schedule provider 'api': undefined method `each' for nil:NilClass`

The current schedules state cannot be fetched, then the schedules are recreated at each Puppet pass. The only clue I have so far is that it comes from commit 13385b7dbad852be906798ccdeb36a6df749cef6

ekohl:
Thanks! That must have slipped through and I'll have a look. The good news is that 5.0 isn't yet released so only git users are affected.

ekohl:
Since I haven't got a test env with schedules yet, could you try to change the get_repos function in util to:
```ruby
def get_repos(type=nil)
  repos = request_api('/v2/repositories/')
  return repos.select { |repo| repo['notes']['_repo-type'] == type } if type
  repos
end
```

ccnifo:
Better, but still an issue : 
`Error: Could not prefetch pulp_schedule provider 'api': undefined method ``name' for #<Array:0x0000000323a070>`

ekohl:
Looks like ```new()``` in ```self.instances``` also broke it. I've submitted a PR to fix it.

PR #261, opened 2017-07-13
Fix pulp_schedule and add tests

ekohl:
In 13385b7dbad852be906798ccdeb36a6df749cef6 I broke the pulp_schedule type. This fixes it, applies some more cleanups and adds tests to prevent future breakages.

Closes GH-260

### DIFF
--- a/lib/puppet/provider/pulp_schedule/api.rb
+++ b/lib/puppet/provider/pulp_schedule/api.rb
@@ -30,50 +30,42 @@ Puppet::Type.type(:pulp_schedule).provide(:api) do
   end
 
   def self.get_resource_properties(repo_id)
-    all = []
-    sync_id = nil
     # get_repo_syncs return an array with all schedules and the repo type
     schedules =  @pulp.get_repo_syncs(repo_id)
-    return all unless schedules[0]
+    return [] unless schedules
+
     all_schedules = schedules[0]
     $schedules_info['repo_type'][repo_id] = schedules[1]
 
     # first get the lowest sync id
-    all_schedules.each { |schedule|
-      if sync_id && sync_id < schedule['_id']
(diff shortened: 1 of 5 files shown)


ccnifo:
I confirm 18a76e38eebcfd754c2b6bb92338d35ba7eb4bc1 fixes GH-260 on my setup, thanks.

ekohl:
@ehelms would you mind having a look?

lib/puppet/provider/pulp_schedule/api.rb
```diff
@@ -30,50 +30,42 @@ def initialize(resource={})
...
         $schedules_info['schedule_id'][schedule['_id']] = repo_id
       end
 
-      hash[:schedule_time] = schedule['schedule']
-      hash[:enabled] = schedule['enabled'] == true ? :true : :false
-      hash[:failure_threshold] = schedule['failure_threshold']
-      all << new(hash)
+      new({
+        :ensure => :present,
+        :provider => :pulp_schedule,
```
alexjfisher: Is this right?  `pulp_schedule` is the type, but the provider is `api`?

ekohl: I will admit that I only copied it from the old code and I'm a bit unsure of what it exactly does.

alexjfisher: and I'll admit I'm only going by what I learnt from http://garylarizza.com/blog/2013/12/15/seriously-what-is-this-provider-doing/
I can't adequately explain why it works/doesn't seem to matter in this case.

ekohl: I merged it now, but maybe we can drop it altogether. I don't really see it being used.

PR #264, opened 2017-08-01
backport "Refactor the class inclusions and chaining #239" to 4.3-stable

evgeni:
required to fix http://projects.theforeman.org/issues/19938 in 3.4.z

### DIFF
--- a/manifests/apache.pp
+++ b/manifests/apache.pp
@@ -86,6 +86,7 @@ class pulp::apache {
       owner   => 'root',
       group   => 'root',
       mode    => '0644',
+      notify  => Service['httpd'],
     }
   } else {
     file {'/etc/httpd/conf.d/pulp.conf':
@@ -94,6 +95,7 @@ class pulp::apache {
       owner   => 'root',
       group   => 'root',
       mode    => '0644',
+      notify  => Service['httpd'],
     }
   }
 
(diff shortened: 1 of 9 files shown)


evgeni:
You mean https://github.com/Katello/katello-installer/commit/1a1754678bf19b5002737b6143bcddf70c370059? We could, but it does not hurt to keep that hook for now, right?

ekohl:
> You mean Katello/katello-installer@1a17546? We could, but it does not hurt to keep that hook for now, right?

It does hurt in my experience. It removes the file in the pre hook on proxies. Sometimes yum installs the file, apache starts and is broken. Only then pulp is configured and this file is created, but since apache failed to start the installer run is broken anyway.

evgeni:
The hook replaces the file, not deletes it, so RPM should never update it.

That's exactly the reason it does that, and not just wipes the file :)

ekohl:
I recall that the file is actually used on the proxy config and I ran into issues with it but I should look into it again to be sure.

ehelms:
This looks like it needs test fixes?

ekohl:
@ehelms for that we need to pin versions in ```.fixtures.yml``` so they are Puppet 3 compatible.

ekohl:
What's the status here?

evgeni:
/me sleeping ;)

let me have another look today

evgeni:
obsolete after https://github.com/Katello/puppet-pulp/commit/59ee6fed7d11cbbf139656bdefd1d006aacba989

PR #266, opened 2017-08-10
Allow newer puppetlabs-concat

ekohl:


ehelms:
Looks like puppet-qpid needs an update?

ekohl:
I'm guessing that it's because puppetlabs-concat 4.0.1 is installed but then puppet-qpid 2.2.0 wants a 2.x version so it breaks. Should we release 2.2.1 which allows concat 4.0?

alexjfisher:
Commit message typo could be confusing?

ekohl:
At least the branch name is correct ...

ekohl:
I tried hacking around it but it looks like the module helper for beaker isn't flexible at all since it doesn't recognize an already installed version.

ekohl:
💚

PR #267, opened 2017-08-11
Allow puppetlabs-mongodb 1.0

ekohl:


ekohl:
https://github.com/Katello/puppet-foreman_proxy_content/pull/135 should be merged as well because librarian-puppet gets confused if module x allows < 2.0 and module y only allows < 1.0 where module x is processed first.

PR #268, opened 2017-08-15
Adding Pulp 2.14 changes

parthaa:
After laboriously going through the links below I identified the 2 files that had their config values updated with the release of pulp-2.14.  one is docker  and the other was python pep. This PR is meant to go in after 2.14 gets tagged to the nightly.

https://github.com/pulp/pulp/compare/2.13-release...2.14-release
https://github.com/pulp/pulp_rpm/compare/2.13-release...2.14-release

https://github.com/pulp/pulp_puppet/compare/2.13-release...2.14-release

https://github.com/pulp/pulp_docker/compare/2.4-release...3.0-release

https://github.com/pulp/pulp_ostree/compare/1.2-release...1.3-release

https://github.com/pulp/crane/compare/2.1-release...3.0-release

https://github.com/pulp/pulp_python/compare/1.0-release...2.0-release

### DIFF
--- a/spec/classes/pulp_apache_spec.rb
+++ b/spec/classes/pulp_apache_spec.rb
@@ -236,6 +236,13 @@ Alias /pulp/docker/v2 /var/www/pub/docker/v2/web
     Options FollowSymlinks Indexes
 </Directory>
 
+<Directory /var/www/pub/docker/v2/web/*/manifests/list>
+    Header set Docker-Distribution-API-Version "registry/2.0"
+    Header set Content-Type "application/vnd.docker.distribution.manifest.list.v2+json"
+    SSLRequireSSL
+    Options FollowSymlinks Indexes
+</Directory>
+
 # Docker v1
 Alias /pulp/docker/v1 /var/www/pub/docker/v1/web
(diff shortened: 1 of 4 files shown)


ehelms:
@parthaa These changes don't appear that they would break a 2.13 or earlier install. If you can confirm that we can pull these now.

ekohl:
And the tests fail on the extra lines. Once that's fixed and no regressions are introduced for 2.13 👍

parthaa:
> @parthaa These changes don't appear that they would break a 2.13 or earlier install. If you can confirm that we can pull these now.

@ehelms Pulp guys confirmed that these changes shouldn't break anything in 2.13 because both config entries specifically refer to files that got added in 2.14

ehelms:
@ekohl should we add a check of Pulp version of these for completion sake? Their existence does not break 2.13 but they are a bit of a misnomer configuration if a user was ever to dig.

ekohl:
@ehelms if it's forwards compatible but dummy in 2.13 I don't mind always including it. You could consider a 2.14+ comment but that doesn't add a lot of value IMHO.

Issue #199, opened 2017-03-07
mongodb user / password not enforced


rgarrigue:
Hi

I tried to use db_user and db_password parameter, which failed. In `/etc/mongod.conf` there's `noauth=true`. And I'ld expect something like thios somewhere in the module

mongodb::db { $db_database:
  user          => $db_username,
  password => $db_password,
}

Here's my log

```
==> test.vagrant.local: Notice: /Stage[main]/Pulp::Database/Exec[migrate_pulp_db]/returns: Attempting to connect to localhost:27017
==> test.vagrant.local: Notice: /Stage[main]/Pulp::Database/Exec[migrate_pulp_db]/returns: Attempting to connect to localhost:27017
==> test.vagrant.local: Notice: /Stage[main]/Pulp::Database/Exec[migrate_pulp_db]/returns: Write concern for Mongo connection: {}
==> test.vagrant.local: Notice: /Stage[main]/Pulp::Database/Exec[migrate_pulp_db]/returns: Database initialization failed: Authentication to MongoDB with username and password failed.
==> test.vagrant.local: Notice: /Stage[main]/Pulp::Database/Exec[migrate_pulp_db]/returns: Authentication to MongoDB with username and password failed.
==> test.vagrant.local: Notice: /Stage[main]/Pulp::Database/Exec[migrate_pulp_db]/returns: Traceback (most recent call last):
==> test.vagrant.local: Notice: /Stage[main]/Pulp::Database/Exec[migrate_pulp_db]/returns:   File "/usr/lib/python2.7/site-packages/pulp/server/db/manage.py", line 198, in main
==> test.vagrant.local: Notice: /Stage[main]/Pulp::Database/Exec[migrate_pulp_db]/returns:     connection.initialize(max_timeout=1)
==> test.vagrant.local: Notice: /Stage[main]/Pulp::Database/Exec[migrate_pulp_db]/returns:   File "/usr/lib/python2.7/site-packages/pulp/server/db/connection.py", line 157, in initialize
==> test.vagrant.local: Notice: /Stage[main]/Pulp::Database/Exec[migrate_pulp_db]/returns:     raise RuntimeError(msg)
==> test.vagrant.local: Notice: /Stage[main]/Pulp::Database/Exec[migrate_pulp_db]/returns: RuntimeError: Authentication to MongoDB with username and password failed.
==> test.vagrant.local: Error: pulp-manage-db && touch /var/lib/pulp/init.flag returned 70 instead of one of [0]
[...]

PR #272, opened 2017-08-31
[WIP] Create the mongodb database with user

ekohl:
Previously we only stored the username/password but didn't actually enforce it. By default mongodb doesn't enforce permissions but if a user configured it to do so, then they would need to manually configure the database.

Currently broken. Needs at least https://github.com/puppetlabs/puppetlabs-mongodb/pull/393 but possibly more.

Closes GH-199


ekohl:
Closing this since I can't work on this and needs fixes in the mongodb module.

PR #274, opened 2017-09-08
crane requires mod wsgi

sean797:


ekohl:
I expected puppetlabs-apache to do this but that's only when ```wsgi_daemon_process`` is set (https://github.com/puppetlabs/puppetlabs-apache/blob/4b46e56e0bd417432c547938dce651a8702a4d66/manifests/vhost.pp#L228-L230). Now I wonder if we should set that instead.

ekohl:
And with https://github.com/puppetlabs/puppetlabs-apache/pull/1693 it'll match the behaviour to my expectations. In the meantime I think we can fix it here as well.

https://modwsgi.readthedocs.io/en/develop/configuration-directives/WSGIDaemonProcess.html states:
> The WSGIDaemonProcess directive can be used to specify that distinct daemon processes should be created to which the running of WSGI applications can be delegated. Where Apache has been started as the root user, the daemon processes can be run as a user different to that which the Apache child processes would normally be run as.

Running as different user could be a good thing (even though we don't do it with pulp's main process either).

sean797:
I think `wsgi_daemon_process` is probably a good thing, but I don't know enough about it, my vote is for merging this as is

ekohl:
That's fair enough as this is a correct solution that touches as little as possible to make it work.

PR #277, opened 2017-09-18
different ca certificate for apache and pulp

timogoebel:
This PR exposes a new parameter `https_ca_cert` to allow you to use a different ca certificate for pulp and apache.

### DIFF
--- a/manifests/apache.pp
+++ b/manifests/apache.pp
@@ -82,7 +82,7 @@ class pulp::apache {
       ssl_cert                   => $::pulp::https_cert,
       ssl_key                    => $::pulp::https_key,
       ssl_chain                  => $::pulp::https_chain,
-      ssl_ca                     => $::pulp::ca_cert,
+      ssl_ca                     => pick($::pulp::https_ca_cert, $::pulp::ca_cert),
       ssl_certs_dir              => '',
       ssl_verify_client          => 'optional',
       ssl_protocol               => $::pulp::ssl_protocol,
--- a/manifests/child/apache.pp
+++ b/manifests/child/apache.pp
@@ -2,7 +2,7 @@ class pulp::child::apache (
   $servername = $::fqdn,
   $ssl_cert = $::pulp::child::ssl_cert,
   $ssl_key = $::pulp::child::ssl_key,
-  $ssl_ca = $::pulp::ca_cert,
+  $ssl_ca = $::pulp::https_ca_cert,
   $max_keep_alive = $::pulp::max_keep_alive,
   $ssl_username = $::pulp::ssl_username,
 ) {
@@ -18,6 +18,12 @@ class pulp::child::apache (
     $directories = undef
   }
 
+  if $ssl_ca {
+    $_ssl_ca = $ssl_ca
+  } else {
+    $_ssl_ca = $::pulp::ca_cert
+  }
+
   apache::vhost { 'pulp-node-ssl':
     servername             => $servername,
     docroot                => '/var/www/html',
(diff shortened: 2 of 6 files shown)


manifests/child/apache.pp
```diff
@@ -18,6 +18,12 @@
     $directories = undef
   }
 
+  if $ssl_ca {
```
ekohl: Can't this block use ```pick()``` as well?

manifests/init.pp
```diff
@@ -396,7 +399,7 @@
     class { '::pulp::crane':
       cert     => $https_cert,
       key      => $https_key,
-      ca_cert  => $ca_cert,
+      ca_cert  => pick_default($https_ca_cert, $ca_cert),
```
ekohl: Why is ```pick_default``` needed here? ```ca_cert``` is already guaranteed to be a valid absolute path.

manifests/apache.pp
```diff
@@ -82,7 +82,7 @@
       ssl_cert                   => $::pulp::https_cert,
       ssl_key                    => $::pulp::https_key,
       ssl_chain                  => $::pulp::https_chain,
-      ssl_ca                     => $::pulp::ca_cert,
+      ssl_ca                     => pick_default($::pulp::https_ca_cert, $::pulp::ca_cert),
```
ekohl: Why is ```pick_default``` needed here?

manifests/init.pp
```diff
@@ -104,6 +104,8 @@
 #
 # $https_chain::                apache chain file for ssl
 #
+# $https_ca_cert::              apache ca cert file for ssl
```
ekohl: Would it make sense to add ```falls back to ca_cert if empty``` or something similar?

manifests/child/apache.pp
timogoebel: Unfortunately not, as pick evaluates all the values and `$::pulp::ca_cert` is not defined in all cases I believe. Does that make sense or am I missing something?

ekohl: Oh yes, the ```pulp::child::apache``` class is weird in its inheritance and something I haven't touched so far. You might be right on this one.

timogoebel:
@ekohl: Updated.

manifests/init.pp
```diff
@@ -294,6 +296,7 @@
   Optional[Stdlib::Absolutepath] $https_cert = $::pulp::params::https_cert,
   Optional[Stdlib::Absolutepath] $https_key = $::pulp::params::https_key,
   Optional[Stdlib::Absolutepath] $https_chain = $::pulp::params::https_chain,
+  Optional[Stdlib::Absolutepath] $https_ca_cert = $::pulp::params::https_ca_cert,
```
sean797: Would it be simpler to default this to `ca_cert` and remove the need for `pick`? Then the option could just be "CA cert for apache" IMO that's simpler for the user.

manifests/init.pp
```diff
@@ -104,6 +104,8 @@
 #
 # $https_chain::                apache chain file for ssl
 #
+# $https_ca_cert::              apache ca cert file for ssl, falls back to ca_cert if empty
```
sean797: ntipick s/empty/undefined/ 
Obviously depending on how you feel about my comment below.

laugmanuel:
@ekohl , @timogoebel, @sean797 given this MR is over 2 years old I would suggest, we close the MR and I open a new one against the current master with the changes implemented as suggested here. Any thoughts?

timogoebel:
Yep, let me close this one as well.

PR #279, opened 2017-10-06
Verify successful bind when creating pulp_rpmbind

### LINKED ISSUE #276: pulp_rpmbind resource creation 'succeeds' for repos that don't exist.
If you use `pulp_rpmbind` to bind to a non-existent repository, the resource should fail to apply.

The provider uses `pulp-consumer`, but the command returns 0 even when it fails.
```
[root@pulp ~]# pulp-consumer rpm bind --repo-id foobar
Repository [foobar] does not exist on the server

[root@pulp ~]# echo $?
0
```
This is with pulp 2.13.2.  Should I try to workaround this in https://github.com/Katello/puppet-pulp/blob/bb68da33dba2e4b4d106ef98aae7b0c9f248f012/lib/puppet/provider/pulp_rpmbind/consumer.rb#L48-L51 ?

alexjfisher:
Solves 2 problems
  * pulp-consumer rpm bind returns 0 even if the repo doesn't exist
  * Binding is asynchronous, so we should wait until it completes

Fixes #276

### DIFF
--- a/lib/puppet/provider/pulp_rpmbind/consumer.rb
+++ b/lib/puppet/provider/pulp_rpmbind/consumer.rb
@@ -46,12 +46,26 @@ Puppet::Type.type(:pulp_rpmbind).provide(:consumer) do
   end
 
   def create
-    consumer('rpm', 'bind', '--repo-id', @resource[:name])
+    output = consumer('rpm', 'bind', '--repo-id', @resource[:name])
+    raise "Repository [#{@resource[:name]}] does not exist on the server" if output =~ /does not exist/
+    wait_for_bind # 'pulp-consumer rpm bind' happens asynchronously in the background
     @property_hash[:ensure] = :present
   end
 
   def destroy
     consumer('rpm', 'unbind', '--repo-id', @resource[:name])
     @property_hash[:ensure] = :absent
+  end
+
+  def wait_for_bind
+    tries ||= 10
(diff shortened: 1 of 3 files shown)


lib/puppet/provider/pulp_rpmbind/consumer.rb
```diff
@@ -46,12 +46,26 @@ def exists?
...
     @property_hash[:ensure] = :present
   end
 
   def destroy
     consumer('rpm', 'unbind', '--repo-id', @resource[:name])
     @property_hash[:ensure] = :absent
   end
+
+  def wait_for_bind
+    tries ||= 10
```
ekohl: I wonder if this should be configurable. What's your experience with how long the waiting typically takes?

alexjfisher: During my testing, the process was actually very quick.  One 250ms wait was *always* enough for me. (IIRC I had to set the sleep to 0.01 to actually see more than one retry take place).  But I figured it might not always be that fast depending on what else the system was doing at the time or because of network congestion etc.  Waiting for no more than 2.5 seconds total seemed reasonable?  If it doesn't complete by then, there's probably no point waiting any longer??

ekohl: That's fair and we can always fix it later.

ekohl:
Thanks!

PR #280, opened 2017-10-11
Fix bad pulp_role example in readme.md

ccnifo:
I'm a bit confused to open a PR for such a small change, but the example path in pulp_role example is wrong: it would silently not work as expected, and may be really misleading.

FYI, a few examples of valid paths for user permissions can be found [here](https://docs.pulpproject.org/dev-guide/integration/rest-api/permission/index.html)


ekohl:
Thanks, these kind of fixes a more than welcome because this is what new users will try. If that fails it's very annoying.

PR #284, opened 2017-10-18
Don't use type defaults for pulp_rpmrepo properties

### LINKED ISSUE #283: pulp_rpmrepo type *properties* should not have defaults
IMO, having property defaults in the type is unhelpful.  For example, if I have an existing repo 'epel7' configured, I should be able to update a single property with something like...
```
puppet resource pulp_rpmrepo epel7 remove_missing=true
```

At the moment, this can also reset a bunch of other properties if you weren't using the default.  (I found this out the hard way when display_name and relative_url were changed.)

Could we move these defaults to the `create` method of the provider?

alexjfisher:
Fixes #283


ekohl:
Thanks!

PR #286, opened 2017-10-25
Refs #21464 - remove pulp oauth support

sean797:


ehelms:
Since base Pulp supports this still in their configuration, we could leave it for now without harm and if Pulp users want this that can use it.

PR #289, opened 2017-11-07
Remove Pulp node support

sean797:
Pulp has deprecated 'Nodes'. http://pulpproject.org/2016/12/07/deprecating-nodes.

Its kind of confusing because we call the Pulp server that runs on ~~Capsules~~ Smart Proxies Pulp Nodes as well, I guess this is left over from when they were actually Pulp Nodes.


ekohl:
While deprecated, I think it's still upstream supported until 3.0. You're right it's confusing though which makes me unsure about removing it. Without knowing our users well enough, it's hard to know if we break their workflows with it.

sean797:
Closing for now

PR #292, opened 2017-11-09
add hook for https vhost includes

rbelnap:
We have a need to provide additional configuration to our ssl vhost for pulp.  This provides a generic hook to specify includes to the vhost.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -104,6 +104,8 @@
 #
 # $https_chain::                apache chain file for ssl
 #
+# $https_includes::             additional includes for https vhost
+#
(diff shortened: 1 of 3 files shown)


manifests/init.pp
```diff
@@ -104,6 +104,8 @@
 #
 # $https_chain::                apache chain file for ssl
 #
+# $https_includes::             additional includes for https vhost
```
ekohl: I'd be good to note that this only applies when ```manage_httpd``` is true.

rbelnap: Oh hey, thanks @ekohl! I did completely miss that, that will work just fine.  You can close this out, unless you have some other use for it (seems doubtful)

ekohl:
Thanks!

PR #293, opened 2017-11-21
Add /pulp/deb to the authority of repo_auth

mdellweg:


ekohl:
Looks like we have an unrelated build issue with the processor count and for some reason it's not running the tests on Puppet 4. This PR itself looks good but I'd like someone with more pulp knowledge to have a look too.

sean797:
FYI looks like https://github.com/puppetlabs/puppetlabs-stdlib/commit/7999ff2aebbd2d85a231318f1e466b36d6dab84e#diff-bb4ff5c3ff02cf9efb5b2d49e312ca39 broke our tests.

sean797:
Restarted travis, they should pass now 🍏

mdellweg:
Seems like we could stabilize on the old tests. New ones will follow.

ekohl:
I've submitted https://github.com/theforeman/puppet-pulp/pull/295 to re-add the puppet 4 tests to the matrix.

manifests/config.pp
```diff
@@ -22,15 +22,28 @@
     target => $pulp::ca_cert,
   }
 
-  if $pulp::enable_rpm {
+  if ($pulp::enable_deb) or ($pulp::enable_ostree) or ($pulp::enable_rpm) {
```
ekohl: The braces are redundant.

manifests/params.pp
```diff
@@ -70,6 +70,7 @@
   $enable_katello = false
   $enable_crane = false
   $enable_rpm = true
+  $enable_deb = true
```
ekohl: If we default to true we should ensure it's present in repositories. At least for Katello this is not true at the time of writing. Not sure about upstream.

templates/etc/httpd/conf.d/pulp_deb.conf.erb
```diff
@@ -0,0 +1,23 @@
+# -- split according to SSL-status
+<Location /pulp/deb/>
+  RewriteEngine On
+  RewriteCond %{HTTPS} on
+  RewriteRule (.+/pulp/deb/)(.*) /pulp/content/var/www/pub/deb/https/repos/$2 [DPI]
```
ekohl: This target path looks rather long. Are you sure this works?

manifests/apache.pp
```diff
@@ -152,6 +152,10 @@
       pulp::apache_plugin { 'rpm': }
     }
 
+    if $::pulp::enable_deb {
+      pulp::apache_plugin { 'deb': }
```
ekohl: Rather than the dummy file you can also use ```vhosts80 => false```.

templates/repo_auth.conf.erb
```diff
@@ -1,11 +1,17 @@
 # This file is managed by puppet. Changes may be overridden!
+<%-
+  repo_url_prefixes = []
+  repo_url_prefixes << "/pulp/deb" if scope['pulp::enable_deb']
+  repo_url_prefixes << "/pulp/repos,/pulp/isos" if scope['pulp::enable_rpm']
```
ekohl: Would you mind splitting this into two lines? Should prepare things for #294

templates/etc/httpd/conf.d/pulp_deb.conf.erb
mdellweg: this is a verbatim copy of the corresponding file in pulp_deb, so i guess yes

.gitignore
```diff
@@ -28,6 +28,9 @@ vendor/
 .rvmrc*
 .ruby-*
 
+## gemset
```
ekohl: This file is modulesynced from https://github.com/theforeman/foreman-installer-modulesync/blob/master/moduleroot/.gitignore.erb. What produces these gemsets?

spec/acceptance/deb_admin_spec.rb
```diff
@@ -0,0 +1,32 @@
+require 'spec_helper_acceptance'
+
+describe 'Scenario: pulp-admin-deb' do
+  let(:pp) do
+    <<-EOS
```
ekohl: It might need (like any non-default plugin) https://github.com/theforeman/puppet-pulp/blob/e25020a7bce87fbec849a20cc0df5f7261600eb8/examples/pulp_puppetrepo.pp#L9-L17 because it's likely another test is running before it - doing the db sync. Then pulp can't do the DB migration when services are running.

spec/acceptance/deb_admin_spec.rb
```diff
@@ -0,0 +1,32 @@
+require 'spec_helper_acceptance'
+
+describe 'Scenario: pulp-admin-deb' do
```
ekohl: This is not only the admin but actually the full deb functionality. That's great but the name should reflect it :)

.gitignore
mdellweg: I did. This is not necessary at all. Will remove it again.

spec/acceptance/deb_admin_spec.rb
mdellweg: OK

Issue #197, opened 2017-02-22
broken update with pulp 2.12


cristifalcas:
With pulp 2.12 it's not possible to update the database anymore if workers and resource manager are running.

Is there any solution for this? It's safe to stop those services before managedb and start them again after?

ehelms:
This should be safe. We stop services before we run managedb in the katello-installer upgrade steps.

ekohl:
In a stand alone situation this is a problem and I'm not sure how to solve this. I think we're forced to use an exec for this.

ekohl:
In an acceptance test I have the following workaround:
```puppet
exec { 'stop services':
  command     => '/bin/systemctl stop pulp_celerybeat pulp_workers pulp_resource_manager pulp_streamer && rm /var/lib/pulp/init.flag',
  subscribe   => Class['pulp::install'],
  before      => Class['pulp::database'],
  refreshonly => true,
}
```

This is not a nice workaround, but it works and is idempotent.

PR #296, opened 2017-12-01
Handle database upgrades by stopping services

ekohl:
Closes GH-197

### DIFF
--- a/examples/pulp_puppetrepo.pp
+++ b/examples/pulp_puppetrepo.pp
@@ -6,16 +6,6 @@ class { '::pulp':
   enable_puppet => true,
 }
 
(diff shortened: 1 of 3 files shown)


ehelms:
Hrmm, when I think about this, I think about the fact that to run `pulp-manage-db` services must be stopped. Seems possibly cleaner to trigger a services stop prior to running `pulp-manage-db` ?

ekohl:
I think that actually makes sense.

ekohl:
https://github.com/theforeman/puppet-pulp/commit/f78dba6d731b11fd9b664aa16d39da62d2404b07 has implemented this.

PR #298, opened 2017-12-12
Fixes #21957 - Pulp-Ostree.conf handle gpg content

parthaa:
Added a Redirect rule to handle requests matching
/pulp/katello/api/repositories/(.*?)/gpg_key_content
The above URL is requested by an Atomic Host
during an rpm-ostree add/remove package operation.

Basically this needs to be redirected to
/katello/api/repositories/$1/gpg_key_content
instead.

### DIFF
--- a/spec/classes/pulp_apache_spec.rb
+++ b/spec/classes/pulp_apache_spec.rb
@@ -348,6 +348,7 @@ Alias /pulp/python /var/www/pub/python/
 #
 RedirectMatch "^/pulp/ostree/web/(.*?)/repodata/(.*)"  "/pulp/repos/$1/repodata/$2"
 RedirectMatch "^/pulp/ostree/web/(.*?)\.rpm"  "/pulp/repos/$1.rpm"
+RedirectMatch "^/pulp/katello/api/repositories/(.*?)/gpg_key_content"  "/katello/api/repositories/$1/gpg_key_content"
 
 # -- HTTPS Repositories ---------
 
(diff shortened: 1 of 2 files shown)


parthaa:
@ekohl @chris1984 not sure why they are failing. Seems somehting more generic than anything related to my template change

chris1984:
Thanks @parthaa

ehelms:
@chris1984 @parthaa @ekohl This is a very specific Katello use case -- shouldn't this have been put into puppet-katello and not in puppet-pulp?

ekohl:
I assumed this was implemented by the pulp-katello plugin rather than katello but you may be right that this might belong somewhere else

ehelms:
Nah - the pulp-katello plugin is really a 'yum clone distributor' and nothing more. This change is hitting an API that is Katello only.

parthaa:
> This is a very specific Katello use case -- shouldn't this have been put into puppet-katello and not in puppet-pulp?
@ehelms 
This is a ostree/katello use case I suppose. I am fine moving it but I donot see puppet_katello mucking with any config files in ->  ```/etc/httpd/conf.d``` (Look at https://github.com/theforeman/puppet-katello/tree/master/templates)
If you have better places to move it to please let us know.

ehelms:
We could put it inside [1] for the server and [2] for a proxy assuming that GPG key content is always accessed over SSL interface?

[1] https://github.com/theforeman/puppet-katello/blob/master/files/pulp-apache-ssl.conf
[2] https://github.com/theforeman/puppet-foreman_proxy_content/blob/master/templates/_pulp_gpg_proxy.erb

ehelms:
@parthaa @ekohl thoughts?

ekohl:
Which client is even accessing this in the first place? Maybe we should fix the clients instead. The whole separation of concerns is hardly present and applications are tightly coupled into a single vhost. I would agree puppet-katello is a better place for the workaround but I'd appreciate if we could at least figure out if clients could be fixed.

ehelms:
@parthaa or @jlsherrill can expand more, but I believe the current situation is a hack due to the way Atomic hosts with ostree work and the fact that Katello is the one who serves up the GPG key even though Pulp serves the content.

PR #299, opened 2018-01-22
Fixes #22338 - add worker_timeout param

chris1984:


manifests/init.pp
```diff
@@ -265,6 +265,10 @@
 #
 # $ldap_remote_user_attribute:: LDAP Remote User Attribute. Defaults to 'sAMAccountName'
 #
+# $worker_timeout::              The amount of time (in seconds) before considering a worker as missing. If Pulp's
```
sean797: nitpick, can you align this to the above text?

PR #307, opened 2018-01-29
Use selboolean

ekohl:
Careful: 100% untested for now.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -88,10 +88,11 @@ class pulp::config {
   }
 
   if $pulp::enable_puppet {
-    exec { 'selinux_pulp_manage_puppet':
-      command => 'semanage boolean -m --on pulp_manage_puppet',
-      path    => '/sbin:/usr/sbin:/bin:/usr/bin',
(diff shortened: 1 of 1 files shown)


manifests/config.pp
```diff
@@ -88,10 +88,11 @@
   }
 
   if $pulp::enable_puppet {
-    exec { 'selinux_pulp_manage_puppet':
-      command => 'semanage boolean -m --on pulp_manage_puppet',
-      path    => '/sbin:/usr/sbin:/bin:/usr/bin',
-      onlyif  => 'getsebool pulp_manage_puppet | grep off',
+    if $::selinux {
+      selboolean { 'pulp_manage_puppet':
```
alexjfisher: Should this use the [defined type](https://github.com/voxpupuli/puppet-selinux/blob/master/manifests/boolean.pp) instead?
```puppet
selinux::boolean{'pulp_manage_puppet':}
```

ekohl: Now that we (indirectly) pull in selinux via squid, we might as well. I do recall that if selinux is disabled the `selboolean` type just fails. I don't see any guards for that and maybe it has been fixed in the base type instead.

PR #308, opened 2018-02-13
Cherry picks for 5.0.1

ekohl:


Gemfile
```diff
@@ -7,6 +7,7 @@ gem 'puppet', ENV.key?('PUPPET_VERSION') ? "~> #{ENV['PUPPET_VERSION']}" : '>= 4
 
 gem 'rake'
 gem 'rspec', '~> 3.0'
+gem 'rdoc', '~> 5.1.0', {"platforms"=>["ruby_21"]}
```
akofink: Why?

ekohl: Current rdoc versions no longer work on Ruby 2.1. Perhaps we could have pinned to < 6.0.0 but this is what we did in all modules and this is just a cherry pick.

PR #310, opened 2018-02-27
Don't specify path to pulp-consumer

alexjfisher:
Specifying the path isn't necessary.
From http://garylarizza.com/blog/2013/11/26/fun-with-providers-part-2/

> We don’t REQUIRE you to pass the full path because sometimes the same
> binary exists in different locations for different operatingsystems.
> Instead of creating a provider for each OS you manage with Puppet, we
> abstract away the path stuff. You CAN still pass a full path as a
> value, but if you elect to do that an the binary doesn’t exist at that
> path, Puppet will disqualify the provider and you’ll be quite upset.

With this simple change I was able to use pulp::consumer on a CentOS 6
system.

### DIFF
--- a/lib/puppet/provider/pulp_rpmbind/consumer.rb
+++ b/lib/puppet/provider/pulp_rpmbind/consumer.rb
@@ -4,7 +4,7 @@ Puppet::Type.type(:pulp_rpmbind).provide(:consumer) do
   desc 'Bind/unbind to an RPM repo'
 
   confine osfamily: :redhat
-  commands consumer: '/bin/pulp-consumer'
+  commands consumer: 'pulp-consumer'
   commands grep:     '/bin/grep'
 
   def self.repo_file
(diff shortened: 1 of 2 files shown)


lib/puppet/provider/pulp_rpmbind/consumer.rb
```diff
@@ -4,7 +4,7 @@
   desc 'Bind/unbind to an RPM repo'
 
   confine osfamily: :redhat
-  commands consumer: '/bin/pulp-consumer'
+  commands consumer: 'pulp-consumer'
   commands grep:     '/bin/grep'
```
ekohl: Should we also change this to just grep?

alexjfisher: You *could* and it wouldn't hurt, but I didn't as it's `/bin/grep` everywhere I think.  Also, I didn't touch the other providers because running a pulp *server* certainly isn't support on EL6.

For anyone interested/googling, with pulp 2.14 repo on CentOS 6.9, I also needed the packages from
* https://copr.fedorainfracloud.org/coprs/g/qpid/qpid/
* https://copr.fedorainfracloud.org/coprs/dgoodwin/subscription-manager/

to meet all the dependencies.

ekohl:
Thanks!

PR #311, opened 2018-02-28
Allow the setting of vhost in consumer messaging section.

glbyers:
When using rabbitmq as a messaging transport for consumer, we should be able to set a vhost. vhost is a valid parameter since 2.8.


glbyers:
Sure. Have added this minimal test case.

ekohl:
puppet-strings 1.2.0 was released and it looks like that changed a few things. I'll look into it.

ekohl:
@glbyers I restarted the jobs now that 1.2.1 includes the regression fix.

glbyers:
np. I think that covers it.

ekohl:
Thanks!

PR #315, opened 2018-04-11
Fixes #23233 - updated yum_max_speed example to reflect to_bytes parsing

chris1984:
Updated the init.pp description too since passing Kb fails with an error, as its the to bytes conversion is not happy with uppercase.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -186,7 +186,7 @@
 #
 # $proxy_password::             Proxy password for authentication
 #
-# $yum_max_speed::              The maximum download speed for a Pulp task, such as a sync. (e.g. "4 Kb" or 1Gb")
+# $yum_max_speed::              The maximum download speed for a Pulp task, such as a sync. (e.g. "4 kb" or "1 Gb")
 #
 # $yum_gpg_sign_repo_metadata:: Whether yum repo metadata GPG signing will be enabled
 #


ekohl:
If pulp understands these strings, why do we even convert them to bytes in the first place? I'd be happy with removing `to_bytes` and enforcing a regex that Pulp understands so we never accept invalid input to begin with.

chris1984:
@ekohl Pulp only understands bytes it was @sean797 who wanted to add the converter in. The converter is the thing that bombs out with Kb instead of kb... do you want to keep it in still since pulp only understands bytes

templates/yum_importer.json.erb
```diff
@@ -20,7 +20,7 @@
     "proxy_password": null,
     <% end %>
     <% unless [nil, :undefined, :undef, ''].include?(scope['pulp::real_yum_max_speed']) -%>
-    "max_speed": <%= scope['pulp::real_yum_max_speed'] %>,
+    "max_speed": "<%= scope['pulp::real_yum_max_speed'] %>",
```
ekohl: If pulp only understands bytes, this change should not be needed.

manifests/init.pp
```diff
@@ -186,7 +186,7 @@
 #
 # $proxy_password::             Proxy password for authentication
 #
-# $yum_max_speed::              The maximum download speed for a Pulp task, such as a sync. (e.g. "4 Kb" or 1Gb")
+# $yum_max_speed::              The maximum download speed for a Pulp task, such as a sync. (e.g. "4 kb" or 1gb")
```
ekohl: to_bytes is weird. `k[bB]` is used, all the others use upper case prefixes: `M[Bb]` `G[Bb]` etc.

chris1984:
@ekohl updated.

chris1984:
@ekohl updated

ekohl:
Thanks!

PR #322, opened 2018-06-13
Generate/regenerate Yum repository metadata GPG signatures

PaulSD:
This is related to https://github.com/theforeman/puppet-certs/pull/188

If a yum repomd GPG key is configured on an existing Pulp server that already has published repositories, then repomd signatures will need to be generated for those existing repositories.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -283,6 +283,9 @@
 # @param yum_gpg_cmd
 #   Custom GPG command/script to use for yum repo metadata signing
 #
+# @param yum_regenerate_repomd_signatures
+#   Whether to regenerate existing yum repo metadata GPG signatures
+#
 # @param num_workers
 #   Number of Pulp workers to use.
 #
@@ -487,6 +490,7 @@ class pulp (
   Boolean $yum_gpg_sign_repo_metadata = $pulp::params::yum_gpg_sign_repo_metadata,
   Optional[String] $yum_gpg_key_id = $pulp::params::yum_gpg_key_id,
   Optional[String] $yum_gpg_cmd = $pulp::params::yum_gpg_cmd,
+  Boolean $yum_regenerate_repomd_signatures = $pulp::params::yum_regenerate_repomd_signatures,
   Integer[0] $num_workers = $pulp::params::num_workers,
   Integer[0] $worker_timeout = $pulp::params::worker_timeout,
   Boolean $enable_admin = $pulp::params::enable_admin,
@@ -567,6 +571,10 @@ class pulp (
 
   Class['pulp::install'] -> Class['pulp::config'] -> Class['pulp::database'] ~> Class['pulp::service', 'pulp::apache']
   Class['pulp::config'] ~> Class['pulp::service', 'pulp::apache']
+
+  if $yum_gpg_sign_repo_metadata {
+    contain ::pulp::repomd_signatures
+  }
 
   if $enable_admin {
     if $ssl_username and $ssl_username != '' {
--- a/manifests/repomd_signatures.pp
+++ b/manifests/repomd_signatures.pp
@@ -1,0 +1,35 @@
+# Generate or regenerate Yum repository metadata GPG signatures
+class pulp::repomd_signatures {
+  # gpg requires $HOME to be set properly, but puppet 'exec' only supports hard-
+  # coding $HOME via the 'environment' parameter (it cannot read it from
+  # /etc/passwd).  To work around this, use `sudo -H` to read /etc/passwd and
+  # set $HOME properly.
+  $shell = "/usr/bin/sudo -u apache -H /usr/bin/bash -c '"
+  $set = '
+    set -e'
+  $loop = '
+    cd /var/lib/pulp/published/yum/master/yum_distributor/ 2>/dev/null || exit 0
+    for d1 in * ; do (
+      cd $d1
+      for d2 in * ; do (
+        cd $d2/repomd
+  '
+  $test = '[ -e repomd.xml.asc ] || '
+  $gen_sig = 'gpg --yes --detach-sign --armor repomd.xml'
+  $stop = 'exit 1'
+  $end = "
+      ) ; done
+    ) ; done
+  '"
+
+  if $::pulp::yum_regenerate_repomd_signatures {
+    exec { 'yum_generate_repomd_signatures':
+      command => "${shell}${loop}${gen_sig}${end}",
+    }
+  } else {
+    exec { 'yum_generate_repomd_signatures':
+      unless  => "${shell}${set}${loop}${test}${stop}${end}",
+      command => "${shell}${loop}${test}${gen_sig}${end}",
+    }
+  }
+}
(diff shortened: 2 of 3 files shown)


manifests/repomd_signatures.pp
```diff
@@ -0,0 +1,12 @@
+# Generate or regenerate Yum repository metadata GPG signatures
+class pulp::repomd_signatures {
+  if $::pulp::yum_regenerate_repomd_signatures {
+    exec { 'yum_regenerate_repomd_signatures':
+      command => "/usr/bin/sudo -u apache -H /usr/bin/bash -c 'cd /var/lib/pulp/published/yum/master/yum_distributor/ 2>/dev/null || exit 0 ; for d1 in `ls` ; do pushd \$d1 ; for d2 in `ls` ; do pushd \$d2/repomd >/dev/null ; gpg --yes --detach-sign --armor repomd.xml ; popd >/dev/null ; done ; popd >/dev/null ; done'",
```
sean797: Please check which attributes would be applicable here https://puppet.com/docs/puppet/5.5/types/exec.html#exec-attributes
At first glance, `user`, `cmd` would probably be helpful.

Also putting it in a multiline string would make it a lot easier to read.

PaulSD: Hmmm...

I can probably use `cwd` here, and I can certainly make it a multiline string.

However, I can't replace the sudo call with a `user` attribute because `gpg` needs `$HOME` to be set properly, but puppet doesn't set `$HOME` (https://projects.puppetlabs.com/issues/5224)...  So I need to use `sudo -H` to ensure that `$HOME` is set properly.

PaulSD:
I don't think these new test failures are caused by my changes...

ekohl:
Could you rebase on master? That should fix the tests.

manifests/init.pp
```diff
@@ -427,6 +430,11 @@
   Class['pulp::install'] -> Class['pulp::config'] -> Class['pulp::database'] ~> Class['pulp::service', 'pulp::apache']
   Class['pulp::config'] ~> Class['pulp::service', 'pulp::apache']
 
+  if $yum_gpg_sign_repo_metadata {
+    include ::pulp::repomd_signatures
+    Class['pulp::repomd_signatures']
```
ekohl: Isn't `contain ::pulp::repomd_signatures` equivalent to this?

manifests/repomd_signatures.pp
```diff
@@ -0,0 +1,21 @@
+# Generate or regenerate Yum repository metadata GPG signatures
+class pulp::repomd_signatures {
+  if !$::pulp::yum_regenerate_repomd_signatures {
```
ekohl: Could you use `unless` instead of `if !`?

manifests/repomd_signatures.pp
```diff
@@ -0,0 +1,21 @@
...
+    $existing_sig_test = '[ -e repomd.xml.asc ] || '
+  }
+  $regen_command = "/usr/bin/sudo -u apache -H /usr/bin/bash -c '
+    cd /var/lib/pulp/published/yum/master/yum_distributor/ 2>/dev/null || exit 0
+    for d1 in `ls` ; do
```
ekohl: Can we use `*` instead of relying on `ls`?

manifests/repomd_signatures.pp
```diff
@@ -0,0 +1,21 @@
...
+      pushd \$d1
+      for d2 in `ls` ; do
+        pushd \$d2/repomd >/dev/null
+        ${existing_sig_test}gpg --yes --detach-sign --armor repomd.xml
```
ekohl: I wonder about `gpg` vs `gpg2`.

manifests/repomd_signatures.pp
```diff
@@ -0,0 +1,21 @@
...
+class pulp::repomd_signatures {
+  if !$::pulp::yum_regenerate_repomd_signatures {
+    $existing_sig_test = '[ -e repomd.xml.asc ] || '
+  }
+  $regen_command = "/usr/bin/sudo -u apache -H /usr/bin/bash -c '
+    cd /var/lib/pulp/published/yum/master/yum_distributor/ 2>/dev/null || exit 0
+    for d1 in `ls` ; do
+      pushd \$d1
+      for d2 in `ls` ; do
+        pushd \$d2/repomd >/dev/null
```
ekohl: Could you use a subshell instead of pushd?

manifests/repomd_signatures.pp
ekohl: You can set HOME via the env parameter of exec.

PaulSD: It is true that you can hard-code a value for HOME via the env parameter of exec.  However, if the user sets a different home directory for the apache user in /etc/passwd (or if the user tries to install foreman on a different Linux distribution that uses a different home directory for apache), then this hard-coded value will be wrong, while `sudo -H` will pick up the correct value from /etc/passwd.

manifests/init.pp
PaulSD: Yes, looks like it is...  Will change.

manifests/repomd_signatures.pp
PaulSD: Ok.

manifests/repomd_signatures.pp
PaulSD: On RHEL7, Ubuntu 18.04, and other current systems, gpg is the same as gpg2, so it shouldn't matter.
On older systems, the command in puppet-pulp should match the command used in pulp_rpm, which is currently `gpg` not `gpg2`:
https://github.com/pulp/pulp_rpm/blob/2.17-release/plugins/pulp_rpm/plugins/distributors/yum/configuration.py#L376

PaulSD:
> This will not be idempotent but I'm not sure I see an easy way to make it so. Perhaps if you extract the existing sig test to a separate script and add that as an `unless` to `exec`.

I'm not quite sure what you mean by this.
This script does nothing if the gpg signature files already exist, so by definition the script is idempotent.
However, this exec will always run even if the script isn't going to do anything, so puppet itself can't determine when changes are actually made by the script.  Is that what you mean by "will not be idempotent"?  If so, then yes, I can add an `unless` to `exec` to fix that problem.

ehelms:
Apologies for it having been a long time since addressing this. I will admittedly be a bit slow cycling on it but I will try to track this better to reduce that cycle time.

Can you explain a bit more about why the installer puppet module needs to handle signing the metadata? I would expect Pulp to handle this rather than having to make a change after Pulp has synced content to a location.

PaulSD:
Pulp normally signs the metadata when publishing a repository:
https://github.com/pulp/pulp_rpm/blob/2.21-release/plugins/pulp_rpm/plugins/distributors/yum/metadata/repomd.py#L49

However, if metadata signing is enabled on an existing server (that already has published repositories), then the installer puppet module needs to generate signatures for those existing repositories.

ekohl:
At this point Katello has removed Pulp 2 and this module is only here to support Katello 3.18. No new features are added to that. That's why I'm closing this. This feature does IMHO belong in Pulp 3, but that's maintained in https://github.com/theforeman/puppet-pulpcore. Apologies for not properly prioritizing this in the Pulp 2 lifecycle.

PR #324, opened 2018-07-09
allow to change http/https port

maxadamo:
Please, allow to change Apache ports through this tiny change. 
In my case I'd like to use Pulp on the same server where I'm already running Nginx, and I'm not willing to configure UWSGI on Nginx.


maxadamo:
disregard. I'm gonna release a fix

PR #325, opened 2018-07-09
allow to change Apache http and https port

### LINKED ISSUE #326: allow changing http and https port
could you please allow to change http/https ports? 
see pull request:  [#325](https://github.com/theforeman/puppet-pulp/pull/325)

**Please Note**: 
2 tests out of 3 were successful, but - unless I'm overlooking - I don't think that the 3rd test has anything to do with my change. To me, it seems that your Travis test is a bit spoiled :-) 
May I suggest you to use PDK?

maxadamo:
Please, allow to change Apache ports through this tiny change.
In my case I'd like to use Pulp on the same server where I'm already running Nginx, and I'm not willing to configure UWSGI on Nginx.

Fixes #326

### DIFF
--- a/manifests/apache.pp
+++ b/manifests/apache.pp
@@ -14,7 +14,7 @@ class pulp::apache {
       apache::vhost { 'pulp-http':
         priority            => '05',
         docroot             => '/usr/share/pulp/wsgi',
-        port                => 80,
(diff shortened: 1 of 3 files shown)


ekohl:
Thanks!

PR #329, opened 2018-07-13
Factor out squid to a separate class

ekohl:


manifests/squid.pp
```diff
@@ -0,0 +1,24 @@
...
+) {
+  class { '::squid3':
+    use_deprecated_opts           => false,
+    http_port                     => [ "${port} accel defaultsite=127.0.0.1:8751" ],
+    acl                           => [ "Safe_ports port ${port}" ],
+    http_access                   => [ 'allow localhost', 'deny to_localhost', 'deny all' ],
+    cache                         => [ 'allow all' ],
+    maximum_object_size           => $maximum_object_size,
+    maximum_object_size_in_memory => $maximum_object_size_in_memory,
+    cache_dir                     => [ 'aufs /var/spool/squid 10000 16 256' ],
```
sean797: Fancy making this configurable while you are at it? Not required.

ekohl: That's kind of the point of this PR. Even if it's only via hiera. The next step would be to switch over to https://github.com/voxpupuli/puppet-squid which is supported. Do you happen to know some good acceptance tests to verify it works?

ekohl:
I'm leaning to closing this in favor of completing https://github.com/theforeman/puppet-pulp/pull/331 instead.

sean797:
I'm in favour of doing #331 instead, but we should try and get an acceptance test working first. I tried the following, it does seem to pass but `/var/log/squid/access.log is empty, so it doesn't look like its actually working.

```
$ cat spec/acceptance/pulp_ondemand_rpmrepo_spec.rb 
require 'spec_helper_acceptance'

describe 'Scenario: pulp_rpmrepo' do
  include_examples 'the example', 'pulp_ondemand_rpmrepo.pp'

  it 'repo syncs' do
    command('pulp-admin rpm repo sync run --repo-id=zoo') do
      its(:exit_status) { is_expected.to eq 0 }
    end
  end

  it 'zoo repo syncs' do
    command('curl http://localhost/pulp/repos/zoo/Packages/w/walrus-5.21-1.noarch.rpm') do
      its(:exit_status) { is_expected.to eq 0 }
    end
  end

end
$ cat examples/pulp_ondemand_rpmrepo.pp 
# The resources in this file serve as examples and are also used in the automated acceptance testing
class { '::pulp':
  manage_repo  => true,
  # https://github.com/theforeman/puppet-pulp/issues/138
  ssl_username => '',
  enable_admin => true,
  manage_squid => true,
}

# But there are several you can set
pulp_rpmrepo { 'zoo':
  display_name      => 'Zoo',
  feed              => 'https://repos.fedorapeople.org/pulp/pulp/demo_repos/zoo/',
  serve_http        => 'TRUE',
  serve_https       => 'TRUE',
  download_policy   => 'on_demand',
}
```

Running it with `BEAKER_destroy=no` and trying to get pulp using squid after just left me with `403` errors, any ideas - maybe I'm missing an param I need to send to `::pulp` but I couldn't see anything.

ehelms:
@ekohl Sounds like all in favor of #331 -- mind if I close this?

PR #331, opened 2018-07-27
Fixes #22746 - Switch to voxpupuli/squid

ekohl:
The module thias/squid3 is unmaintained so we're switching to a maintained module. We also take the chance to make it possible to override some parameters.

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -6,7 +6,7 @@ fixtures:
     concat:  "https://github.com/puppetlabs/puppetlabs-concat.git"
     mongodb: "https://github.com/voxpupuli/puppet-mongodb.git"
     qpid:    "https://github.com/theforeman/puppet-qpid.git"
-    squid3:  "https://github.com/thias/puppet-squid3.git"
+    squid:   "https://github.com/voxpupuli/puppet-squid.git"
     systemd: 'https://github.com/camptocamp/puppet-systemd.git'
     yumrepo_core:
       repo: "https://github.com/puppetlabs/puppetlabs-yumrepo_core"
--- a/manifests/squid.pp
+++ b/manifests/squid.pp
@@ -1,0 +1,55 @@
+# The class to manage squid. This is used by pulp streamer.
+class pulp::squid(
+  Stdlib::Port $port = 3128,
+  Stdlib::Host $streamer_host = '127.0.0.1',
+  Stdlib::Port $streamer_port = 8751,
+  String $maximum_object_size = '5 GB',
+  String $maximum_object_size_in_memory = '100 MB',
+  Stdlib::Absolutepath $cache_dir = '/var/spool/squid',
+) {
+  class { 'squid':
+    maximum_object_size_in_memory => $maximum_object_size_in_memory,
+  }
+
+  squid::acl { 'Safe_ports':
+    type    => 'port',
+    entries => [$port],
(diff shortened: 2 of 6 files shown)


ekohl:
This is untested for now.

.fixtures.yml
```diff
@@ -6,6 +6,6 @@ fixtures:
     concat:  "https://github.com/puppetlabs/puppetlabs-concat.git"
     mongodb: "https://github.com/voxpupuli/puppet-mongodb.git"
     qpid:    "https://github.com/theforeman/puppet-qpid.git"
-    squid3:  "https://github.com/thias/puppet-squid3.git"
+    squid:   "https://github.com/voxpupuli/puppet-squid.git"
   symlinks:
```
alexjfisher: Take the opportunity to remove the redundant `symlinks` section?

ekohl: We should probably do this on all our modules at the same time to keep them consistent.

ehelms:
Needs a rebase, but overall looks good. I'm happy to test this once updated.

ekohl:
Rebased. I've added a very minimal acceptance test but I'd like suggestions on a more functional test.

ehelms:
The functional aspect is more about the streamers interaction with squid, which feels like it would be difficult to do a full acceptance test of. I think what you have here is good.

ekohl:
Looks like the squid module doesn't accept `100 MB` and the type is way too limiting. Going to submit a patch for that.

ekohl:
https://github.com/voxpupuli/puppet-squid/pull/112. We'll also need a release for the acceptance tests to pass since otherwise there's an issue with using stdlib 5.x.

ekohl:
We'll first need a squid module release with the new data type before acceptance tests can pass and this can be merged. If this hasn't happened soon, I'll do it myself. Until then this is WIP.

ekohl:
The build is green and I think it should be ready but I haven't gotten to actual real world tests of this.

manifests/squid.pp
```diff
@@ -0,0 +1,55 @@
+# The class to manage squid. This is used by pulp streamer.
+class pulp::squid(
+  Integer[0, 65535] $port = 3128,
```
alexjfisher: I'm guessing this PR might have predated `Stdlib::Port`?

ekohl: At least predates me being aware/using it.

ehelms:
Thanks @ekohl !

alexjfisher:
Built a new pulp server yesterday using this and no issues.

ekohl:
Only downside is that we now need to drop Puppet 4 support in the installer: first module to require Puppet >= 5.5.8.

alexjfisher:
I'm not going to object if you want to release a puppet/squid 1.2.0...

ekohl:
I'm pretty sure I needed a fix that was in 2.0.0 and I don't see a lot of objections to dropping 4.10 so not going to bother :)

alexjfisher:
There aren't any fixes on the master branch after the 'drop puppet 4' commit, so branching just before that commit would work.  But if you don't need puppet 4 support, then you'll be keeping @bastelfreak happy by not making a final 1.x release! :)

bastelfreak:
Please drop Puppet 4 since  it's end of life since weeks.

PR #337, opened 2018-09-05
Remove redundant symlink

ekohl:
With recent versions of puppetlabs_spec_helper this is done automatically


sean797:
Fancy updating the Gemfile to require minimum `v2.10.0` ?

ekohl:
How so? it was added in 1.2.0 according to https://github.com/puppetlabs/puppetlabs_spec_helper/blob/master/CHANGELOG.md#120---2016-08-23

PR #338, opened 2018-09-06
Add max_speed to iso_importer

sean797:


templates/iso_importer.json.erb
```diff
@@ -19,4 +19,9 @@
     <% else %>
     "proxy_password": null
```
ekohl: This needs a trailing comma now

templates/iso_importer.json.erb
```diff
@@ -19,4 +19,9 @@
...
     <% end %>
+    <% unless [nil, :undefined, :undef, ''].include?(scope['pulp::real_yum_max_speed']) -%>
+    "max_speed": <%= scope['pulp::real_yum_max_speed'] %>,
```
ekohl: Doesn't need a trailing comma. Same below

ekohl:
I'm wondering whether we should use https://github.com/puppetlabs/puppetlabs-stdlib#to_json_pretty for these files.

sean797:
Thanks for the quick review.. +1 to to_json_pretty in another PR

ekohl:
I do wonder if it's obvious that the `yum_max_speed` parameter also counts for the ISO importer.

sean797:
I wouldn't say its clear that it counts for RPM either ;)

Updated the doc  :+1:

sean797:
Ugh, just realised what you were saying.. I'm not too keen on having `iso_max_speed` as well, and renaming it is a pain :cry:

ehelms:
Given these live in the same plugin, I am OK with this one parameter for now.

PR #339, opened 2018-09-14
Allow newest module dependencies

ekohl:


metadata.json
```diff
@@ -10,7 +10,7 @@
   "dependencies": [
     {
       "name": "puppetlabs/stdlib",
-      "version_requirement": ">= 4.6.0 < 5.0.0"
+      "version_requirement": ">= 4.6.0 < 6.0.0"
```
alexjfisher: @ekohl How would you feel about bumping the minimum up to say [4.25.0](https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/CHANGELOG.md#supported-release-4250) so that we could use some of the newer types such as `Stdlib::Port`?

ekohl: I don't mind that but I'd prefer that bumping to be part of the PR that starts to use those data types.

alexjfisher: Ok.  I've gone ahead and updated https://github.com/theforeman/puppet-pulp/pull/341 ;)

Issue #340, opened 2018-09-18
Apache vhost for crane uses wrong certificate


alexjfisher:
Given the following...
```puppet
class { 'pulp':
    # ...
    https_cert     => '/etc/httpd/ssl/example.com.crt',
    https_key      => '/etc/httpd/ssl/example.com.key',
    https_chain    => '/etc/httpd/ssl/geotrust-int.crt',
    enable_crane   => $enable_crane,
}
```

I would expect the `05-pulp-https.conf` vhost and the `03-crane.conf` vhosts to use identical https certificate chains.  But... `03-crane.conf` sets `SSLCertificateChainFile` to `/etc/pki/pulp/ca.crt` instead of `/etc/httpd/ssl/geotrust-int.crt` (which `05-pulp-https.conf` correctly uses).

PR #341, opened 2018-09-19
Make crane use `$pulp::https_chain` if specified

alexjfisher:
Currently when declared from `pulp.pp`, crane sets its vhost `ssl_chain`
parameter to `$pulp::ca_cert` instead of `$pulp::https_chain`.

Fixes #GH-340

Since `pulp::crane` might be used independently of `pulp`, I've tried to
make this bug fix backwards compatible for those users.


manifests/crane.pp
```diff
@@ -20,17 +22,23 @@
 #
 # $ssl_protocol::               SSLProtocol configuration to use
 class pulp::crane (
+  Stdlib::Absolutepath $key,
+  Stdlib::Absolutepath $cert,
+  Stdlib::Absolutepath $ca_cert,
```
alexjfisher: The parameters aren't `Optional`, so I don't see how it makes sense to default them to `undef` (via params.pp)

manifests/crane.pp
```diff
@@ -20,17 +22,23 @@
...
-  Stdlib::Absolutepath $ca_cert = $::pulp::crane::params::ca_cert,
+  Optional[Stdlib::Absolutepath] $ssl_chain = $::pulp::crane::params::ssl_chain,
   Optional[String] $ssl_protocol = $::pulp::crane::params::ssl_protocol,
 ) inherits pulp::crane::params {
-  class { '::pulp::crane::install': } ~>
-  class { '::pulp::crane::config': } ~>
-  class { '::pulp::crane::apache': } ->
-  Class['pulp::crane']
+
+  contain pulp::crane::install
```
alexjfisher: I think `contain` is more expressive than the pattern this is replacing.

manifests/crane/apache.pp
```diff
@@ -1,11 +1,18 @@
...
-  include ::apache::mod::proxy_http
-  include ::apache::mod::wsgi
+  include apache
+  include apache::mod::headers
+  include apache::mod::proxy
+  include apache::mod::proxy_http
+  include apache::mod::wsgi
+
+  if $::pulp::crane::ssl_chain {
+    $ssl_chain = $::pulp::crane::ssl_chain
```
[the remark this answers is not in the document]
alexjfisher: Yes, of course.  Somehow completely forgot that function existed!

manifests/crane/apache.pp
```diff
@@ -1,11 +1,18 @@
 # Sets up Apache for Crane
 class pulp::crane::apache {
 
-  include ::apache
-  include ::apache::mod::headers
-  include ::apache::mod::proxy
-  include ::apache::mod::proxy_http
-  include ::apache::mod::wsgi
+  include apache
```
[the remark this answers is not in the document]
alexjfisher: OK.  I'll do it all in a single PR later.

manifests/crane.pp
```diff
@@ -20,17 +22,23 @@
...
-) inherits pulp::crane::params {
-  class { '::pulp::crane::install': } ~>
-  class { '::pulp::crane::config': } ~>
-  class { '::pulp::crane::apache': } ->
-  Class['pulp::crane']
+  Stdlib::Absolutepath $key,
+  Stdlib::Absolutepath $cert,
+  Stdlib::Absolutepath $ca_cert,
+
+  Boolean                        $debug                     = false,
```
ekohl: I've stopped caring about alignment because I think it creates too much noise in changes. With Puppet 4 types it can also hinder readability instead of add to it IMHO.

alexjfisher: Alignment removed.

alexjfisher:
@ekohl Tests are failing as we need a new release of `puppet/mongodb` following https://github.com/voxpupuli/puppet-mongodb/pull/484 being merged?

PR #343, opened 2018-10-16
don't set a default lazy_redirect_host

evgeni:
this allows Pulp to decide itself which host to redirect to

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -86,7 +86,7 @@ class pulp::params {
   $email_enabled = false
 
   $manage_squid = false
-  $lazy_redirect_host = downcase($::fqdn)
+  $lazy_redirect_host = undef
   $lazy_redirect_port = undef
   $lazy_redirect_path = '/streamer/'
   $lazy_https_retrieval = false


evgeni:
Maybe, this is not as easy as I thought:

```
% git grep redirect_ho      
server/etc/pulp/server.conf:# redirect_host:
server/etc/pulp/server.conf:# redirect_host:
server/pulp/server/config.py:        'redirect_host': socket.getfqdn(),
server/pulp/server/content/web/views.py:        redirect_host = pulp_conf.get('lazy', 'redirect_host')
server/pulp/server/content/web/views.py:            redirect_host or host,
server/pulp/server/controllers/repository.py:    host = pulp_conf.get('lazy', 'redirect_host')
server/test/unit/server/content/web/test_views.py:                'redirect_host': host,
server/test/unit/server/controllers/test_repository.py:            'redirect_host': 'pulp.example.com',
```

ekohl:
On IRC we discussed that setting it to a space and a colleague confirmed that did what he wanted. That results in a YAML that is valid but passes an empty string into the app I think. It's a rather ugly hack though.

ehelms:
Can you expand on why you want to make this change?

evgeni:
@ehelms the relevant pulp patch was merged: https://github.com/pulp/pulp/pull/3726 - do we want to merge this now, or wait till a pulp release contains it, or smth else? :)

ehelms:
With the Pulp patch, what will this change cause to happen? As in, will things operate the same if we pull this in town?

evgeni:
So the pulp patch switches the default redirect host from `socket.getfqdn()` to whatever was in the HTTP Host header.

ekohl:
Can this be released into 1.20 or will it need to wait for that Pulp patch to actually be released?

evgeni:
@ekohl see my previous comment, I think this should be safe for 1.20 without the Pulp change.

PR #344, opened 2018-10-23
Add remove_missing to pulp_deb

mdellweg:
We want to default to mirroring mode in debian.


templates/deb_importer.json.erb
```diff
@@ -1,4 +1,5 @@
 {
+    "remove_missing": true,
```
ekohl: Should this be configurable?

mdellweg: Not if i can avoid it. ;)
See also:
https://github.com/Katello/katello/pull/7773

PR #350, opened 2018-11-21
Move vhosts80 to httpd conf.d

ehelms:


ekohl:
I was wondering about this. Is this something coming from Pulp itself or is it purely something we do in our module?

spec/classes/pulp_apache_spec.rb
```diff
@@ -22,7 +22,7 @@
       is_expected.to contain_class('apache::mod::ssl')
     end
 
-    it { is_expected.to contain_file('/etc/pulp/vhosts80/')}
+    it { is_expected.to contain_file('/etc/httpd/conf.d//vhosts80/')}
```
ekohl: The double slash looks like a problem

manifests/apache.pp
```diff
@@ -17,7 +17,7 @@
         port                => $::pulp::http_port,
         servername          => $::fqdn,
         serveraliases       => [$::hostname],
-        additional_includes => '/etc/pulp/vhosts80/*.conf',
+        additional_includes => '/etc/httpd/conf.d/vhosts80/*.conf',
```
ekohl: Could you store this base directory in a variable (as we should have in the beginning) and use `$apache::confd_dir` in it? That way other modules can just refer to that variable rather than copying it.

ehelms:
The Pulp RPM handles creating this directory in `/etc/pulp` and each plugin populates it. This has been that way for a long time, since Pulp 2 had no official configuration or installation other than RPM. So this moves it away from what the RPM does to more centralized in Apache.

ehelms:
@ekohl Do you recognize the error? It seems like its unrelated to the change

ekohl:
Since you're including Apache, you need facts in the tests. The easiest way is here:

https://github.com/theforeman/puppet-pulp/blob/3b9b634c00774294e08afeb3914b800f501799fc/spec/defines/pulp_child_fragment_spec.rb#L13-L16

ehelms:
@ekohl Thanks for all the pointers, this is now passing

evgeni:
The diff looks correct and would achieve the intended goal. I'd like to do some name bikeshedding though.

As an admin, I'd wonder what this `/etc/httpd/conf.d/vhosts80/` folder is for. Should we name it `pulp-vhosts80` or something similar, to indicate it's pulp-related?

manifests/apache.pp
```diff
@@ -139,7 +139,7 @@
   if $pulp::manage_httpd or $pulp::manage_plugins_httpd {
     pulp::apache_plugin {'content' : vhosts80 => false}
 
-    file { '/etc/pulp/vhosts80/':
+    file { "${$apache::confd_dir}/vhosts80/":
```
evgeni: missing `pulp-` prefix

evgeni: and the `$` before `apache` is missplaced

PR #358, opened 2019-02-21
Refs #26103 - Make the crane server name configurable

ekohl:


manifests/crane.pp
```diff
@@ -27,6 +27,7 @@
   Stdlib::Absolutepath $ca_cert,
 
   Boolean $debug = false,
+  Stdlib::Fqdn $server_name = $::fqdn,
```
ekohl: Kafo has had issues with `$facts['fqdn']` so here I'm intentionally using the older version

PR #360, opened 2019-02-28
Improve pulp_rpmrepo gpg_key validation and docs

alexjfisher:


lib/puppet/type/pulp_rpmrepo.rb
```diff
@@ -187,7 +187,14 @@ def munge_boolean_to_symbol(value)
   end
 
   newproperty(:gpg_key) do
-    desc "GPG key used to sign and verify packages in the repository"
+    desc <<-DESC
+File containing public GPG keys used to validate the signatures of RPMs and metadata in this repository.
```
alexjfisher: Taken from https://docs.pulpproject.org/plugins/pulp_rpm/tech-reference/yum-plugins.html#optional-configuration-parameters

ekohl:
Thanks!

PR #361, opened 2019-02-28
Add `scheduled_repo` wrapper defined type

alexjfisher:
I never create an `pulp_rpmrepo` without also creating a `pulp_schedule`.  This wrapper provides a convenient shortcut.

### DIFF
--- a/functions/daily_schedule.pp
+++ b/functions/daily_schedule.pp
@@ -1,0 +1,10 @@
+# @api private
+# @param seed A seed to pass to `fqdn_rand`.  A good choice is the repo name that this schedule will be for.
+# @return Returns a 'pseudo-random' daily iso8601 schedule.
+# @example Calling the function
+#    pulp::daily_schedule('centos-7')
+function pulp::daily_schedule(String $seed) >> String {
+  $random_hour   = sprintf('%02d', fqdn_rand(23, $seed))
+  $random_minute = sprintf('%02d', fqdn_rand(59, $seed))
+  "2000-01-01T${random_hour}:${random_minute}Z/P1D"
+}
--- a/manifests/scheduled_repo.pp
+++ b/manifests/scheduled_repo.pp
@@ -1,0 +1,36 @@
+# @summary A wrapper that creates a pulp_Xrepo resource with and an associated pulp_schedule resource
+#
+# @example Basic usage with default `rpm` `repo_type` and 'random' daily schedule
+#   pulp::scheduled_repo { 'centos-7':
+#     repo_config => {
+#       'display_name' => 'CentOS 7 Base Repo',
+#       'feed'         => 'https://www.mirrorservice.org/sites/mirror.centos.org/7/os/x86_64',
+#     },
+#   }
+#
+# @param repo_type
+#   Determines whether a `pulp_rpmrepo`, `pulp_puppetrepo` or `pulp_isorepo` resource is created.
+# @param repo_config
+#   Defines a hash of parameters to pass to the `pulp_Xrepo` resource.
+# @param repo_schedule
+#   An iso8601 schedule string or the special value 'daily'. Defaults to 'daily' which creates a `pulp_schedule` with a pseudo-random daily `schedule_time`.
+define pulp::scheduled_repo
+(
+  Enum['rpm','puppet','iso'] $repo_type = 'rpm',
+  Hash $repo_config = {},
+  Variant[Enum['daily'],Pulp::Iso8601TimeInterval] $repo_schedule = 'daily',
+)
+{
+  $schedule_time = $repo_schedule ? {
+    'daily' => pulp::daily_schedule($name),
+    default => $repo_schedule,
+  }
+
+  Resource["pulp_${repo_type}repo"] { $name:
+    * => $repo_config,
+  }
+
+  pulp_schedule { $name:
+    schedule_time => $schedule_time,
+  }
+}
--- a/types/iso8601timeinterval.pp
+++ b/types/iso8601timeinterval.pp
@@ -1,0 +1,3 @@
+# https://en.wikipedia.org/wiki/ISO_8601#Time_intervals
+# TODO: Actually implement a regex.
+type Pulp::Iso8601TimeInterval = String[1]
--- a/spec/defines/pulp_scheduled_repo_spec.rb
+++ b/spec/defines/pulp_scheduled_repo_spec.rb
@@ -1,0 +1,87 @@
+require 'spec_helper'
+
+describe 'pulp::scheduled_repo' do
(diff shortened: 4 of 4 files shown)


alexjfisher:
@ekohl If the basic idea looks ok to you, I'll add some spec tests.

ekohl:
I'm not opposed to this. Our native types are already a combination of various resources. https://github.com/theforeman/puppet-pulp/issues/253 was intended to split this: create simple native types that straight up map to the API and then combine them in a higher level. Never had the time to finish it.

functions/daily_schedule.pp
```diff
@@ -0,0 +1,9 @@
+# @param seed A seed to pass to `fqdn_rand`.  A good choice is the repo name that this schedule will be for.
+# @return Returns a 'pseudo-random' daily iso8601 schedule.
+# @example Calling the function
+#    pulp::daily_schedule('centos-7')
+function pulp::daily_schedule(String $seed) >> String {
```
alexjfisher: I've actually been using this function for a long time in my profile module.  Thought it was time to move it somewhere more appropriate. :)

manifests/scheduled_repo.pp
```diff
@@ -0,0 +1,38 @@
...
+#   Determines whether a `pulp_rpmrepo`, `pulp_puppetrepo` or `pulp_isorepo` resource is created.
+# @param repo_config
+#   Defines a hash of parameters to pass to the `pulp_Xrepo` resource.
+# @param schedule
+#   An iso8601 schedule string or the special value 'daily'. Defaults to 'daily' which creates a `pulp_schedule` with a pseudo-random daily `schedule_time`.
+define pulp::scheduled_repo
+(
+  Enum['rpm','puppet','iso'] $repo_type = 'rpm',
+  Hash $repo_config = {},
+  Variant[Enum['daily'],String[1]] $schedule = 'daily',
```
alexjfisher: I appreciate `String[1]` already encompasses `'daily'`.  Hopefully this documents the intent though and one day someone might replace `String[1]` with an actual regex pattern to use for an iso8601 interval.

I suppose I could create a type alias (even if the implementation sucks).

manifests/scheduled_repo.pp
```diff
@@ -0,0 +1,38 @@
...
+)
+{
+  $schedule_time = $schedule ? {
+    'daily' => pulp::daily_schedule($name),
+    default => $schedule,
+  }
+
+  case $repo_type { # lint:ignore:case_without_default
```
alexjfisher: a `default` would be unreachable since `$repo_type` is an `Enum`

ekohl: An alternative that might work:
```puppet
"pulp_${repo_type}repo" { $name:
  * => $repo_config,
}
```

alexjfisher: ooh.  I didn't think this would work.  Now that I've written some tests, it's easy to find out.

alexjfisher: `Error while evaluating a Resource Statement, Evaluation Error: Illegal Resource Type expression, expected result to be a type name, or untitled Resource, got String`

alexjfisher:
Renamed parameter `schedule` to `repo_schedule` as `schedule` is already the name of a metaparameter.

manifests/scheduled_repo.pp
ekohl: I guess `create_resource` could be used and I think here it's actually cleaner

alexjfisher: Really? I thought that was a left over from the puppet 3 days and considered evil.

alexjfisher: Evil to some people anyway...
https://puppet.com/docs/puppet/3.8/future_lang_resources_advanced.html#implementing-the-createresources-function
https://www.craigdunn.org/2016/04/designing-modules-in-a-puppet-4-world

ekohl: But in those cases the resource type is not variable :)

binford2k: This looks weird, but should work. (eyeball compiled only)

```
Resource["pulp_${repo_type}repo"] { $name:
  * => $repo_config,
}
```

binford2k: I should clarify that I'm not yet convinced that I like this pattern yet. It's... not exactly intuitively readable unless you use mental gymnastics like this:

> index into all *resources* to find the type  "pulp_${repo_type}repo"

ekohl: I wonder this would be easier to read:
```puppet
$resource_type = Resource["pulp_${repo_type}repo"]
$resource_type { $name:
  * => $repo_config,
}
```

Feels a bit redundant, but my mind has trouble with the uppercase `Resource`. Makes me think it refers to an existing instance rather than defining a new one.

alexjfisher: Easier to read, but no longer valid. :)
`Evaluation Error: Illegal Resource Type expression, expected result to be a type name, or untitled Resource, got Type[Pulp_rpmrepo]`

@binford2k's example does work though.

types/iso8601timeinterval.pp
```diff
@@ -0,0 +1,3 @@
+# https://en.wikipedia.org/wiki/ISO_8601#Time_intervals
+# TODO: Actually implement a regex.
+type Pulp::Iso8601TimeInterval = String[1]
```
alexjfisher: *Maybe* the proper pattern would be something like...
```
^(?:[\+-]?\d{4}(?!\d{2}\b))(?:(-?)(?:(?:0[1-9]|1[0-2])(?:\1(?:[12]\d|0[1-9]|3[01]))?|W(?:[0-4]\d|5[0-2])(?:-?[1-7])?|(?:00[1-9]|0[1-9]\d|[12]\d{2}|3(?:[0-5]\d|6[1-6])))(?:[T\s](?:(?:(?:[01]\d|2[0-3])(?:(:?)[0-5]\d)?|24\:?00)(?:[\.,]\d+(?!:))?)?(?:\2[0-5]\d(?:[\.,]\d+)?)?(?:[zZ]|(?:[\+-])(?:[01]\d|2[0-3]):?(?:[0-5]\d)?)?)?)?\/(-?)P(?=\d|T\d)(?:(\d+)Y)?(?:(\d+)M)?(?:(\d+)([DW]))?(?:T(?:(\d+)H)?(?:(\d+)M)?(?:(\d+(?:\.\d+)?)S)?)?$
```

alexjfisher: But I haven't got the motivation to actually check that works! :)

alexjfisher:
Updated with `@api private` for function.

PR #362, opened 2019-03-12
Fixes #26323 - Remove max_tasks_per_child setting

chris1984:
Will open up Refs prs to the other modules that have this setting including katello-devel so they can be merged after this one is accepted.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -200,9 +200,6 @@
 #
 # $enable_crane::               Whether to enable crane docker repository
 #
-# $max_tasks_per_child::        Number of tasks after which the worker is restarted and the memory it allocated is
-#                               returned to the system
-#
 # $enable_rpm::                 Whether to enable rpm plugin.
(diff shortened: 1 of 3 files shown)


ehelms:
Was this setting removed from Pulp or have they deprecated it's use? If not, I think we need to keep this here and only remove it form modules that use this module.

chris1984:
Proxy pr:
https://github.com/theforeman/puppet-foreman_proxy_content/pull/196

chris1984:
@ehelms the feature is still there just not being moved over with Pulp 3. I am fine with that suggestion on just removing it from our stuff.

ekohl:
I agree with @ehelms that as long as it's a feature, it's fine to keep it in. We don't need to expose it as a top level parameter though.

chris1984:
Closing based on discussions, will just keep the katello/proxy-content one open and make a migration for the installer.

PR #364, opened 2019-04-11
Drop Puppet 4

ekohl:


ehelms:
The failure looks unrelated but I am surprised to see it.

ekohl:
The squid module pulls in puppet-selinux. Not sure how I think about that.

evgeni:
I think adding selinux to fixtures is correct, as we do not parse fixture module deps otherwise

PR #367, opened 2019-04-18
Convert documentation to yardoc

ekohl:
This converts the documentation from RDoc to Yardoc. The benefit is that puppet-strings can read it which allows easy generation of reference documentation (`rake strings:generate:reference`). The only benefit of RDoc over Yardoc is support for groups which the installer can use to hide advanced sections, but this module isn't exposed by the installer and didn't have any groups.

For the 6.0.0 release I did this locally which results in https://forge.puppet.com/katello/pulp/reference. There are still some issues:
* `<` and `>` are not escaped which means various examples aren't very usable.
* Longer term it also makes sense to move static defaults from ::params to the main class so it can statically read the default.
* Some native types lack documentation

### DIFF
--- a/manifests/admin.pp
+++ b/manifests/admin.pp
@@ -1,77 +1,106 @@
-#
-# == Class: pulp::admin
-#
 # Install and configure Pulp admin
 #
-# === Parameters:
+# @param version
+#   Pulp admin package version, it's passed to ensure parameter of package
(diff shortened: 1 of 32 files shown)


alexjfisher:
Any reason behind not using `@summary` tag?

ekohl:
Lazyness? This just converted the parameters and got a mostly usable reference. As I said: there are still issues :)

PR #369, opened 2019-05-16
Update gofer package names for EL8

ehelms:


evgeni:
EL8 is not a supported release for this module ;)

https://github.com/theforeman/puppet-pulp/blob/cad1cd0f0b2c58ab3e2ed90e256e00849e0c79d2/metadata.json#L46-L65

evgeni:
(also probably should have tests)

ehelms:
Updated

PR #370, opened 2019-05-21
Allow puppet/mongodb 3.x

ekohl:


ehelms:
I think this may have broken some logic (and I have yet to find out where). This now appears in the installer:

```
    Error: rh-mongodb34-mongodb-syspaths conflicts with mongodb-2.6.12-6.el7.x86_64
```

Something inside the mongo module appears to be installing mongodb from EPEL despite -- https://github.com/theforeman/foreman-installer/blob/develop/config/foreman.hiera/family/RedHat-7.yaml

PR #372, opened 2019-06-05
Added passthrough pulp mode

sbernhard:
This change implements a passthrough mode on a pulpnode (=capsule) to pulp master using the already existing squid proxy.

The advantage is, that many customers don't want to duplicate the pulp content but still have separated subnets. Therefore, we configured a squid proxy on the capsule to route all requests for pulp content (rpm, apt) to pulp master (=katello)

For using this at katello: 
- install the katello-ca-consumer-latest.noarch.rpm of the pulp master on the content host (instead of the capsule ca consumer)
- register via the subscription manager
- set the proxy in yum.conf OR in apt.conf to use the proxy like
   proxy=http://capsule.dev:3129


ehelms:
This is an interesting idea. I have a few questions to help me think on it.

 1) This couples the squid deployed to help with lazy content syncing to a squid used for passthrough proxying. What if I want one but not the other?
 2) In puppet-foreman_proxy_content, a reverse proxy is deployed through Apache that in theory could route back to the main server. What is the advantage of squid in this case? Given it requires using a separate port as well?
 3) How does one deploy this configuration to a proxy without deploying all of Pulp given it seems tightly coupled to puppet-pulp?

sbernhard:
Thanks Eric for you question and your interest. 
To 1: I added the passthrough mode as a addition - means, the pulp streamer scenario for lazy content syncing should still work. 
To 2: I guess, because of SSL certificates to have a secure connection AND to authenticate on pulp (using the client cert) a reverse proxy can not be used. Additionally, I prefered to have the squid solution as its possible to verify that only requests of a given network to pulp master is allowed. Don't know if this is possible with apache. 
3. it is coupled with pulp. I think, it might be possible to use passthrough pulp without having the rest of the pulp scenario but this is MUCH more effort and I guess, the katello-agent can not be used as it still requires qpid and (I think) some pulp services (am I right?)

ehelms:
I think I am not grasping the concept here. Let me recap my thoughts and you can steer me back on course. I interpreted this as a lightweight way to present content on a proxy so that hosts isolated could get content from the main server without having to have a full blown Pulp configured and syncing content.

sbernhard:
This is correct.

During installation, all pulp node services are still configured. If you want, you can still sync the content to the pulp node (capsule) - maybe only parts. In the end, the client decides if the passthrough mode is used (= proxy configuration on the hosts is configured) or the pulp node.

Of course, it doesn't make sense to sync all the content to the pulp node and then use the passthrough mode. The intention is to use the passthrough mode and don't sync the content so that the content (rpms, debs) doesn't need to be duplicated.


---- Eric Helms schrieb ----


I think I am not grasping the concept here. Let me recap my thoughts and you can steer me back on course. I interpreted this as a lightweight way to present content on a proxy so that hosts isolated could get content from the main server without having to have a full blown Pulp configured and syncing content.

sbernhard:
ping @ehelms

sbernhard:
ping @ehelms / @ekohl

manifests/init.pp
```diff
@@ -517,6 +533,11 @@
   Optional[String] $ldap_bind_dn = $pulp::params::ldap_bind_dn,
   Optional[String] $ldap_bind_password = $pulp::params::ldap_bind_password,
   String $ldap_remote_user_attribute = $pulp::params::ldap_remote_user_attribute,
+  Boolean $enable_passthrough_pulp = $pulp::params::enable_passthrough_pulp,
+  Stdlib::Port $passthrough_pulp_http_port = $pulp::params::passthrough_pulp_http_port,
+  Optional[String] $passthrough_pulp_allowed_net = $pulp::params::passthrough_pulp_allowed_net,
```
ekohl: You can enforce the data type with https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/REFERENCE.md#stdlibipaddressv4cidr. You may want to consider https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/REFERENCE.md#stdlibipaddressv6cidr as well.

manifests/squid.pp
```diff
@@ -47,7 +71,7 @@
     order          => '60',
     config_entries => {
       'cache_peer'          => "${streamer_host} parent ${streamer_port} 0 no-digest no-query originserver name=PulpStreamer",
-      'cache_peer_access'   => 'PulpStreamer allow all',
+      'cache_peer_access'   => 'PulpStreamer allow localhost',
```
ekohl: This looks like a general thing that can be applied to all installations, correct? Only Apache/Pulp connect to squid.

manifests/init.pp
sbernhard: @ekohl See https://github.com/theforeman/puppet-pulp/pull/372#issuecomment-499354560 regarding your answer about "full pulp instance deployment".

manifests/squid.pp
sbernhard: yes. I think so.

sbernhard:
can you please have a look at it again @ekohl ?

ehelms:
@sbernhard We have had a hard time wrapping our brains around this change and how it plays into the architecture. With Pulp 3 integration appearing in Katello releases, and these components no longer being present. Is this a change you still want to persue?

PR #378, opened 2019-11-22
Drop default ssl_protocol and pass it to crane

ekohl:
When enabling crane from the main class, the certificates are already matched. When the user has passed ssl_protocol to tighten the ciphers, they can also be passed to crane.

This also changes the default to undef which means the Apache default is used. This allows hardening the entire server without setting it for all separate vhosts.


evgeni:
```
expected that the catalogue would contain Apache::Vhost[pulp-https] with ssl_protocol set to ["all", "-SSLv2", "-SSLv3"] but it is set to [nil]
```

;-)

ekohl:
Updated and :green_apple: now

PR #381, opened 2019-11-29
Add support for different CA certificates for apache and pulp

laugmanuel:
This PR allows the configuration of a different CA cert for the Apache Vhost and introduces the new parameter `https_ca_cert`. 

It's a follow up to: https://github.com/theforeman/puppet-pulp/pull/277

### DIFF
--- a/manifests/apache.pp
+++ b/manifests/apache.pp
@@ -81,7 +81,7 @@ class pulp::apache {
       ssl_cert                   => $pulp::https_cert,
       ssl_key                    => $pulp::https_key,
       ssl_chain                  => $pulp::https_chain,
-      ssl_ca                     => $pulp::ca_cert,
+      ssl_ca                     => $pulp::https_ca_cert,
       ssl_certs_dir              => '',
       ssl_verify_client          => 'optional',
       ssl_protocol               => $pulp::ssl_protocol,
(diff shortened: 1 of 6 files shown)


ekohl:
I'll let the tests continue before merging but otherwise it looks good to merge. I've been known to forget checking back so feel free to ping if it looks like that's the case.

laugmanuel:
Tests are green - I think we are ready for merge if there are no other thoughts @ekohl

ekohl:
Thanks!

evgeni:
It seems this broke Katello? https://ci.centos.org/job/foreman-katello-nightly-test/653/console

```
systemd[1]: Starting The Apache HTTP Server...
httpd[22387]: AH00526: Syntax error on line 62 of /etc/httpd/conf.d/05-pulp-https.conf:
httpd[22387]: SSLCACertificateFile: file '/etc/pki/pulp/ca.crt' does not exist or is empty
systemd[1]: httpd.service: main process exited, code=exited, status=1/FAILURE
```

timogoebel:
@evgeni: @laugmanuel is on PTO today.

evgeni:
@timogoebel then I guess @ekohl has to jump in ;)

laugmanuel:
@evgeni this change did indeed break Katello when a custom `ca_cert` is specified. I did implement a fix and it is currently running tests. I will open a PR as soon as the local tests are green.
Sorry for the trouble!

laugmanuel:
For reference https://github.com/theforeman/puppet-pulp/pull/382

PR #382, opened 2019-12-09
Fix https_ca_cert if a different location for ca_cert is used

laugmanuel:
This fixes the bug introduced here https://github.com/theforeman/puppet-pulp/pull/381 which occurs if the default path of `pulp::ca_cert` is changed. 
The root cause is that `pulp::https_ca_cert` is set to the value of `pulp::ca_cert` in `params.pp`. If the value of `pulp::ca_cert` gets changed by passing an argument to the class, we have a missmatch if `pulp::https_ca_cert` is not defined at the same time.

@ekohl do you mind taking a look at the fix?
@evgeni FYI

### DIFF
--- a/manifests/apache.pp
+++ b/manifests/apache.pp
@@ -81,7 +81,7 @@ class pulp::apache {
       ssl_cert                   => $pulp::https_cert,
       ssl_key                    => $pulp::https_key,
       ssl_chain                  => $pulp::https_chain,
-      ssl_ca                     => $pulp::https_ca_cert,
+      ssl_ca                     => pick($pulp::https_ca_cert, $pulp::ca_cert),
       ssl_certs_dir              => '',
       ssl_verify_client          => 'optional',
       ssl_protocol               => $pulp::ssl_protocol,
--- a/manifests/child/apache.pp
+++ b/manifests/child/apache.pp
@@ -20,6 +20,14 @@ class pulp::child::apache (
     $directories = undef
   }
 
+  if $ssl_ca {
+    $_ssl_ca = $ssl_ca
+  } elsif $::pulp::ca_cert {
+    $_ssl_ca = $::pulp::ca_cert
+  } else {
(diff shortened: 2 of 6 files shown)


evgeni:
thanks @laugmanuel!

manifests/child/apache.pp
```diff
@@ -31,7 +39,7 @@
     ssl                    => true,
     ssl_cert               => $ssl_cert,
     ssl_key                => $ssl_key,
-    ssl_ca                 => $ssl_ca,
+    ssl_ca                 => $_ssl_ca,
```
ekohl: Doesn't this also work:
```suggestion
    ssl_ca                 => pick($ssl_ca, $::pulp::ca_cert, $pulp::child::server_ca_cert),
```

manifests/apache.pp
```diff
@@ -81,7 +81,7 @@
       ssl_cert                   => $pulp::https_cert,
       ssl_key                    => $pulp::https_key,
       ssl_chain                  => $pulp::https_chain,
-      ssl_ca                     => $pulp::https_ca_cert,
+      ssl_ca                     => pick($pulp::https_ca_cert, $pulp::ca_cert),
```
ekohl: Given the repetition of this, I'm wondering if in `init.pp` there should be a `$real_https_ca_cert`. Both are ugly.

manifests/child/apache.pp
laugmanuel: If I remember correctly there was one scenario which broke the tests if I did it that way.

manifests/apache.pp
laugmanuel: We could do that. But I suspect we need the workaround in `manifests/child/apache.pp` anyway because of some special inheritance scenario. I would prefer to keep the logic close to the location where the variable is used.

ekohl:
Thanks!

PR #383, opened 2019-12-11
Handle V2 API via Pulp 2 WSGi explicitly

ehelms:
This helps when running Pulp 2 and Pulp 3 on the same system given
they share a path root of /pulp/api


ehelms:
For now I think it's OK (and we'd have to check if thats being used or not) to set the environment variable no matter which pulp API it talks to.

ekohl:
Reverted in https://github.com/theforeman/puppet-pulp/pull/384 so skipping in the changelog.

PR #385, opened 2019-12-13
Move /etc/httpd/conf.d/{,10-}pulp.conf

wbclark:
This is to resolve https://github.com/theforeman/puppet-pulpcore/issues/49 by moving `pulp.conf` to `10-pulp.conf` so it loads after the Foreman vhost.

### DIFF
--- a/manifests/apache.pp
+++ b/manifests/apache.pp
@@ -125,7 +125,7 @@ class pulp::apache {
       notify  => Service['httpd'],
     }
   } else {
-    file {'/etc/httpd/conf.d/pulp.conf':
+    file {'/etc/httpd/conf.d/10-pulp.conf':
       ensure  => file,
       content => template('pulp/pulp.conf.erb'),
       owner   => 'root',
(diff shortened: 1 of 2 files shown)


manifests/apache.pp
```diff
@@ -116,7 +116,7 @@
     }
 
     # This file is installed by pulp-server but we have everything in the above vhost
-    file {'/etc/httpd/conf.d/pulp.conf':
+    file {'/etc/httpd/conf.d/10-pulp.conf':
```
ekohl: This won't work because (as the comment above states) is created by the pulp-server RPM.

ehelms: Ahh right, so we can leave it be as it was.

wbclark: Will https://github.com/theforeman/puppet-pulpcore/pull/50 alone be sufficient?

ehelms:
We still need the changes you have below renaming the file. This one being
inside a pure Pulp deployment doesn't need to get renumbered.

wbclark:
I've removed the renaming on line 119

ekohl:
I still think merging this is incorrect because of https://github.com/theforeman/puppet-pulp/pull/385#discussion_r357853524. You'll just end up with 2 config files now.

PR #390, opened 2020-02-18
Fixes #29054: Run pulp migrations if any are pending

ehelms:


manifests/database.pp
```diff
@@ -9,11 +9,11 @@
...
 
   exec { 'migrate_pulp_db':
-    command   => 'pulp-manage-db && touch /var/lib/pulp/init.flag',
+    command   => 'pulp-manage-db',
     path      => '/bin:/usr/bin',
     logoutput => 'on_failure',
     user      => 'apache',
-    creates   => '/var/lib/pulp/init.flag',
     timeout   => $pulp::migrate_db_timeout,
+    unless    => '/bin/pulp-manage-db --dry-run',
```
ekohl: Probably better to leave out the path and keep it the same as in `command`.
```suggestion
    unless    => 'pulp-manage-db --dry-run',
```

PR #391, opened 2020-02-19
Refs #29052: Stop services before running database migrations

ehelms:


ehelms:
Requires https://github.com/theforeman/puppet-pulp/pull/390 and updates to puppetlabs-transition

ehelms:
Rebased on the PR that handles checking for pending migrations.

manifests/database.pp
```diff
@@ -8,12 +8,34 @@
     Service['mongodb'] -> Exec['migrate_pulp_db']
   }
 
+  transition { 'stop pulp_workers':
```
ekohl: How about:
```puppet
['workers', 'streamer', 'celerybeat', 'resource_manager'].each |$service| {
  transition { "stop pulp_${service}":
    resource   => Service["pulp_${service}"],
    attributes => { ensure => 'stopped' },
    prior_to   => Exec['migrate_pulp_db'],
  }

  Exec['migrate_pulp_db'] ~> Service["pulp_${service}"]
}
```

I'm not entirely sure the refresh is needed since init.pp has it, but I left it in the example here just to show how it can be done with less code duplication.
https://github.com/theforeman/puppet-pulp/blob/1300985bfa981738afbd9f9895581d8753d65b80/manifests/init.pp#L573

ehelms:
Updated with something based on your suggestion

ehelms:
I'm good with waiting a bit. Thanks for tackling the transitions module part. I think it will find other uses as well. Pulp 3 will need a similar treatment from me as they have the same restraint around having services stopped before making database changes.

I'd be curious if Candlepin or Foreman should as well.

PR #394, opened 2020-04-08
Drop redundant modulesync config

ekohl:
I don't know if this actually works, but the tests will show this.


ekohl:
This module certainly needs EPEL.

PR #396, opened 2020-04-17
Modulesync: use voxpupuli-test 1.2

ekohl:


ehelms:
Has some errors that aren't clear to me

spec/classes/pulp_config_spec.rb
```diff
@@ -1,42 +1,29 @@
 require 'spec_helper'
-
+require 'deep_merge'
 
 describe 'pulp::config' do
-  let :default_facts do
-    on_supported_os['redhat-7-x86_64'].merge(:processorcount => 3)
+  let :facts do
+    {processors: {'count' => 3}}.deep_merge(on_supported_os['redhat-7-x86_64'])
```
ekohl, line 6: This looks weird and is IMHO not intuitive, but it matches https://github.com/danielsdeleo/deep_merge/blob/9b15cc77c5954eebf67365d4f24ca44a9e2b2ca7/test/test_deep_merge.rb#L144-L148 so I assume it'll continue to work.

ekohl:
:green_heart: now

PR #399, opened 2020-05-04
Support puppet/extlib 5.x

ehelms:


metadata.json
```diff
@@ -14,7 +14,7 @@
     },
     {
       "name": "puppet/extlib",
-      "version_requirement": ">= 3.0.0 < 5.0.0"
+      "version_requirement": ">= 5.0.0 < 6.0.0"
```
ekohl, line 17: There's no reason to increase the lower bound. Doing so only limits us and we're perfectly compatible with the entire range.
```suggestion
      "version_requirement": ">= 3.0.0 < 6.0.0"
```

ehelms: I don't get it. For EL8 support, we have to have 5.X version. So I would have thought we have to raise that lower bound to ensure proper working support of this module across the supported OSes?

ekohl: Technically this module doesn't support EL8 at all and probably never will. However, in the larger environment we still want to pull in extlib 5.0 so everything has to be compatible. Similar to when you want to pull in a major version of a new gem.

ehelms: Sorry, still confused. The 5.0.0 version introduced marking support for EL8. To me that says, this module only guarantees EL8 support on 5.X and above.

I suppose you are saying, they code on all versions will work on EL8 we've just "marked" it that way. And are allowing lower versions because we know it will work. Which I get in a way, but seems strange to me since the whole point of marking support and semantic versioning is to lock into versions that have indicated support for a feature.

I'll lower it back down, just another area of the puppet ecosystem that seems confusing and doesn't always follow the rules.

ekohl: This module works with versions 3.x - 5.x. EL8 support is irrelevant for this module since Pulp 2 doesn't run on EL8.

If it would support EL8, it's a more complicated situation because there are no general guidelines on how to deal with it and I'll agree with you on that. On the one hand, you can argue for open since there are valid combinations (user doesn't run EL8) and leave it up to the end user who assembles it (this is the approach I've mostly taken in the past). You can also argue for closed (to support EL8, all dependencies are tightened to versions that work with EL8).

AFAIK there is no dependency solver that takes OS support into account and it's mostly a hint to the admin who picks version support. Often a new OS doesn't need actual code changes in modules. Extlib for example doesn't have manifests and only has Puppet functions. You could argue OS support is actually irrelevant https://github.com/puppetlabs/puppetlabs-transition takes that approach by not listing individual OS versions but it's not standardized.

alexjfisher: > Technically this module doesn't support EL8 at all and probably never will. However, in the larger environment we still want to pull in extlib 5.0 so everything has to be compatible. Similar to when you want to pull in a major version of a new gem.

Of course the functions run on the puppetserver, which isn't necessarily on the same server running pulp, so this is pretty moot too.

PR #400, opened 2020-05-11
Refs #29715: Add options to expose mongodb on a public interface with…

ehelms:
… SSL

### DIFF
--- a/spec/classes/pulp_database_spec.rb
+++ b/spec/classes/pulp_database_spec.rb
@@ -1,0 +1,62 @@
+require 'spec_helper'
+
+describe 'pulp::database' do
+  let :facts do
+    on_supported_os['redhat-7-x86_64']
+  end
+
+  context 'with default parameters' do
+    it 'should configure mongodb::server for localhost' do
+      is_expected.to contain_class('mongodb::server').with({
+        'bind_ip' => ['127.0.0.1'],
+        'ipv6'    => false,
+        'ssl'     => '',
+        'ssl_key' => '',
+        'ssl_ca'  => '',
+      })
+    end
+  end
+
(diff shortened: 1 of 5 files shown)


spec/classes/pulp_database_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+
+    it 'should configure mongodb::server for localhost' do
+      is_expected.to contain_class('mongodb::server').with({
+        'bind_ip' => ['0.0.0.0'],
+        'ipv6'    => false,
+        'ssl'     => true,
+        'ssl_key' => '/tmp/cert-bundle.key',
+        'ssl_ca'  => '/tmp/ca.crt',
+      })
+    end
```
ehelms, line 38: Locally this fails with:

```
  3) pulp::database with mongodb parameters and ipv6 should configure mongodb::server for localhost
     Failure/Error:
       is_expected.to contain_class('mongodb::server').with({
         'bind_ip' => ['0.0.0.0', '::/0'],
         'ipv6'    => true,
         'ssl'     => true,
         'ssl_key' => '/tmp/cert.key',
         'ssl_ca'  => '/tmp/ca_cert.crt',
       })
     
     Puppet::PreformattedError:
       Evaluation Error: Resource type not found: Stdlib::Absolutepath (file: /home/ehelms/workspace/upstream/modules/puppet-pulp/spec/fixtures/modules/pulp/manifests/init.pp, line: 417, column: 3) on node war.eagle
```

@ekohl can you see what I am doing wrong?

ekohl: This can happen if you didn't prepare the fixtures (`rake spec_prep`) and manually ran `rspec`.

PR #403, opened 2020-06-01
Improve validation of ldap parameters

alexjfisher:
Make sure that none of the ldap parameters are `undef` if any of them
have been set.

With this pattern, (using `NotUndef`), we avoid repeating the data types
already specified in the parameter list.


alexjfisher:
This PR obviously isn't that important and is more an example that anything else. (How to implement conditional parameter validation was discussed on slack this morning, and the benefits of using `NotUndef` came up)

PR #410, opened 2020-11-02
Switch from Travis CI to Github Actions

ekohl:


ehelms (approved):
This is the same current breakage of master we see with Travis which shouldn't hold up merging this change.

ekohl:
Looks like MongoDB was removed from EPEL7. We don't notice this elsewhere because we switch to the RH SCL for it.

ehelms:
Let's fix that separately from this change since we can see it here for example https://github.com/theforeman/puppet-pulp/pull/409

PR #415, opened 2021-07-08
Fixes #32966 - add remove_old_repodata option for yum

jlsherrill:


spec/classes/pulp_config_spec.rb
```diff
@@ -72,6 +72,52 @@
...
+
+    end
+  end
+
+  context "with nil remove_old_repodata configuration" do
+    let :pre_condition do
+      "class {'pulp':
+        enable_rpm                    => true,
+        yum_remove_old_repodata_count => undef
+      }"
```
jlsherrill, line 104: Setting to undef here doesn't actually work, it defaults to the default.     Is there another strategy to do this?  Some thoughts:
should I have 2 params?  one to enable, and one to set the count, similar to the pulp config

jlsherrill:
note that this is likely for a downstream only release

manifests/init.pp
```diff
@@ -484,6 +487,7 @@
   Boolean $yum_gpg_sign_repo_metadata = $pulp::params::yum_gpg_sign_repo_metadata,
   Optional[String] $yum_gpg_key_id = $pulp::params::yum_gpg_key_id,
   Optional[String] $yum_gpg_cmd = $pulp::params::yum_gpg_cmd,
+  Optional[Integer[0,65535]] $yum_remove_old_repodata_count = $pulp::params::yum_remove_old_repodata_count,
```
ehelms, line 490: Since this is optional, set the default value in `params.pp` to `undef`.

templates/yum_distributor.json.erb
```diff
@@ -0,0 +1,8 @@
+{
+    <% unless [nil, :undefined, :undef, ''].include?(scope['pulp::yum_remove_old_repodata_count']) -%>
```
ehelms, line 2:
```suggestion
    <% if scope['pulp::yum_remove_old_repodata_count'] -%>
```

spec/classes/pulp_config_spec.rb
ehelms: I think this will solve it: https://github.com/theforeman/puppet-pulp/pull/415/files#r666904364

templates/yum_distributor.json.erb
```diff
@@ -0,0 +1,8 @@
...
+    "remove_old_repodata": True,
+    "remove_old_repodata_threshold": <%= scope['pulp::yum_remove_old_repodata_count'] %>
+    <% else %>
+    "remove_old_repodata": False
```
ekohl, line 6: is `False` valid JSON?

manifests/config.pp
```diff
@@ -59,6 +59,15 @@
       mode      => '0644',
       show_diff => $pulp::show_conf_diff,
     }
+
+    file { '/etc/pulp/server/plugins.conf.d/yum_distributor.json':
+      ensure    => file,
+      content   => template('pulp/yum_distributor.json.erb'),
```
ekohl, line 65: Other than a template, you can also use [to_pretty_json](https://forge.puppet.com/modules/puppetlabs/stdlib/reference#to_json_pretty).

manifests/init.pp
jlsherrill: i was trying to make this be a one commit chnage, rather than also having to modify (downstream only) puppet-katello and puppet-foreman_proxy_content and i think if the default is unset then i'd have to do that.  Which is fine, i just want to clarify that is kinda your intent too

ehelms: Drop the `Optional` and set the default value to 0. I think that gets you what you want the easiest.

templates/yum_distributor.json.erb
jlsherrill: good point, i was going by what was mentioned in this bug: https://bugzilla.redhat.com/show_bug.cgi?id=1903367  

but it was later corrected, should be true

jlsherrill:
simplified this quite a bit, as we don't care about this being configurable, and this will be easier to get into downstream

manifests/init.pp
jlsherrill: done!  As per our discussion this would allow it to be configured via heira if needed, and set back to the code default of 4

manifests/config.pp
jlsherrill: I like that! but i think i'm just gonna follow the existing convention here if thats okay

ehelms (approved):
Tests passed locally, not sure why not on the PR itself

